PR #21, opened 2019-07-25
Search with pagination

hedgehog1833:
yeha, first one \o/
general search and pagination


src/main/java/com/metalr2/config/mvc/WebConfig.java
```diff
@@ -11,7 +11,7 @@
 
   @Override
   public void addViewControllers(ViewControllerRegistry registry) {
-    registry.addViewController(Endpoints.Frontend.FOLLOW_ARTISTS).setViewName(ViewNames.Frontend.FOLLOW_ARTISTS);
+//    registry.addViewController(Endpoints.Frontend.FOLLOW_ARTISTS).setViewName(ViewNames.Frontend.FOLLOW_ARTISTS);
```
Waginator: remove comment

src/main/java/com/metalr2/web/controller/SearchController.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+@Controller
+@Slf4j
+public class SearchController {
```
Waginator: the name of controller should be FollowArtistsController

src/main/java/com/metalr2/web/controller/SearchController.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @ModelAttribute
+  private SearchRequest searchRequest() {
+    return new SearchRequest();
+  }
+
+  @PostMapping({Endpoints.Frontend.FOLLOW_ARTISTS})
+  public ModelAndView handleSearchRequest(@ModelAttribute SearchRequest searchRequest) {
+    log.info(searchRequest.getArtistName());
```
Waginator: Logging should be more descriptive here. If the application is running productively, it is not very meaningful if only the name of the artists appears in the log file. Log-Level is in my opinion DEBUG or TRACE.

Furthermore: If the log output gets values from a method call, then the following procedure is best practice: log.debug("{}", searchRequest.getArtistName());

Reason: The method is now called only if the log level of the application is DEBUG. The reason for this is performance. Possibly there is an expensive operation behind the method call.

src/main/java/com/metalr2/web/controller/SearchController.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @GetMapping({Endpoints.Frontend.FOLLOW_ARTISTS})
+  public ModelAndView showSearch(@RequestParam(name = "size", defaultValue = DEFAULT_PAGE_SIZE) int size,
+                                 @RequestParam(name = "page", defaultValue = DEFAULT_PAGE) int page,
+                                 @RequestParam(name = "artistName", defaultValue = DEFAULT_ARTIST_NAME) String artistName) {
+    if (artistName.isEmpty()) {
+      return getDefaultModelAndView();
+    }
+
+    Optional<ArtistSearchResults> artistSearchResultsOptional = artistSearchRestClient.searchForArtist(artistName, String.valueOf(page), String.valueOf(size));
```
Waginator: Please compare lines 49 to 60 and 71 to 81. I think you can extract this into a method.

src/main/java/com/metalr2/web/controller/SearchController.java
```diff
@@ -0,0 +1,117 @@
...
+    return new ModelAndView(ViewNames.Frontend.FOLLOW_ARTISTS, viewModel);
+  }
+
+  private Map<String,Object> buildViewModel(ArtistSearchResults artistSearchResults, String artistName) {
+    Map<String, Object> viewModel = new HashMap<>();
+    viewModel.put("artistName", artistName);
+    viewModel.put("artistSearchResultList", artistSearchResults.getResults());
+
+    Pagination pagination = artistSearchResults.getPagination();
+    List<String> pageNumbers = IntStream.rangeClosed(1, pagination.getPagesTotal()).boxed().map(String::valueOf).collect(Collectors.toList());
```
Waginator: is transformation to String really necessary?

src/main/java/com/metalr2/web/controller/SearchController.java
```diff
@@ -0,0 +1,117 @@
...
+    viewModel.put("currentPage", pagination.getCurrentPage());
+    viewModel.put("pageNumbers", pageNumbers);
+
+    PaginationUrls paginationUrls = pagination.getUrls();
+    String nextSize = DEFAULT_PAGE_SIZE;
+    String nextPage = DEFAULT_PAGE;
+
+    if (paginationUrls.getNext() != null){
+      nextSize = String.valueOf(pagination.getItemsPerPage());
+      nextPage = String.valueOf(pagination.getCurrentPage()+1);
```
Waginator: I think you don't need to convert the values into a String

src/main/java/com/metalr2/web/dto/request/SearchRequest.java
```diff
@@ -0,0 +1,15 @@
...
+
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class SearchRequest {
```
Waginator: the name should be ArtistSearchRequest

src/main/resources/static/css/frontend.css
```diff
@@ -47,3 +47,28 @@ footer{
...
+.btn-pagination {
+    margin-right: 0.5rem;
+    background-color: #ececec;
+    color: #212529;
+    padding: 0.25rem;
+}
+
+.btn-pagination:hover {
+    background-color: #777777;
+}
```
Waginator: add new line

src/main/resources/templates/frontend/follow-artists.html
```diff
@@ -1,6 +1,55 @@
 <!DOCTYPE html>
 <html data-th-replace="~{frontend/fragments/layout :: layout (~{::body})}" lang="en">
-  <body>
+<body>
```
Waginator: The body and the first div container are not correctly indented...

Should be:
```
<html>
  <body>
    <div>
    ...
    </div>
  </body>
</html>
```

src/main/java/com/metalr2/web/controller/SearchController.java
hedgehog1833: Not there specifically, that's correct. But there are some places where I need the String and some where I need an int.
I tried to improve it somewhat but it's still kind of weird, if you don't want to hold both types independantly

src/main/resources/templates/frontend/follow-artists.html
hedgehog1833: IntelliJ suggested this for me, but I guess your version looks better

Issue #22, opened 2019-07-25
Improvements for follow artists page


hedgehog1833:
Currently it is possible send an empty string which results in all (?) artists on discogs and the pagination exploding to 400 pages. Sending an empty string should be prevented and result in a message to the user.
The same should happen if there are no results for the given term.
Replace hardcoded discogs-link in follow-artsts.html

PR #30, opened 2019-07-27
Issue#23 npe page non existing

hedgehog1833:
Yeah I think using the Optional is nicer too. So I changed a few things, here are my comments:
- The ArtistSearchRestClient validates the input parameters artistName and size. Is the name empty or the size 0 an empty Optional is instantly provided. An empty Optional is also provided if the searchResult's body is null, the HTTP status is not OK or the resultList is empty. All these cases result in an output to the user that his input parameters did not produce any results.
- All user input parameters are logged on debug level. Errors like the 404 are logged as ERROR by the CustomClientErrorHandler anyway so we don't need to do this too.
- The assertion in the controller was proposed by IntelliJ: I already checked that the searchResults are not null in die RestClient. The controller does not know this but putting an assertion there stops IntelliJ to complain :)

EDIT: this also solved all issues mentioned in #22 Improvements for follow artists page


src/main/resources/static/css/frontend.css
```diff
@@ -72,3 +72,7 @@ footer{
 .btn-pagination:hover {
     background-color: #777777;
 }
+
```
Waginator: Please only overwrite the bootstrap CSS classes if you can clearly exclude side effects. The class .container is still used in many places and the lower distance pushes the layout together a bit. You can see it on the other pages. Better use your own class for this or work with spacing classes from bootstrap (https://getbootstrap.com/docs/4.3/utilities/spacing/). Just user class mb-3 as css class and you have a margin-bottom of 1rem.

src/main/java/com/metalr2/web/controller/FollowArtistsController.java
```diff
@@ -44,23 +45,24 @@ private ArtistSearchRequest searchRequest() {
...
   }
 
   @GetMapping({Endpoints.Frontend.FOLLOW_ARTISTS})
   public ModelAndView showFollowArtists(@RequestParam(name = "size", defaultValue = DEFAULT_PAGE_SIZE) int size,
                                         @RequestParam(name = "page", defaultValue = DEFAULT_PAGE) int page,
                                         @RequestParam(name = "artistName", defaultValue = DEFAULT_ARTIST_NAME) String artistName) {
-    if (artistName.isEmpty()) {
-      return getDefaultModelAndView();
+    if (artistName.equals(DEFAULT_ARTIST_NAME) && size == Integer.parseInt(DEFAULT_PAGE_SIZE)
+                                               && page == Integer.parseInt(DEFAULT_PAGE)){
```
Waginator: I know what you mean, but it's hard to understand...Maybe we need a better indicator to differentiate between requesting the search form and requesting the next or previous page of a result set. For now, I have no idea. We should fix it later. I'll add it to my notes.

PR #40, opened 2019-08-09
Issue#16 artist details

hedgehog1833:
Added a simple artist details page. It's not nice that all images are loaded at once but as we are going to change the layout anyway I thought for a first try it may be okay this way.


src/main/java/com/metalr2/web/controller/ArtistDetailsController.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private ModelAndView createArtistDetailsModelAndView(String artistName, long artistId) {
+    Optional<Artist> artistOptional = artistSearchRestClient.searchForArtistById(artistId);
+
+    if (artistOptional.isEmpty()) {
+      return createBadArtistSearchRequestModelAndView(artistName, artistId);
+    }
+
+    Artist artist = artistOptional.get();
```
Waginator: Line 46 to 63: 
The two ArtistController handle the ViewModel very differently.

Very often only single values are put into the view map. This makes the controller methods very extensive. 
However, there are also cases where DTOs of the Discogs API are simply passed on to the client. If the Discogs API changes, this would probably also result in changes to the client.

Solution:
- For the responses of the Discogs API own DTOs are used, which are not used for the HTML templates. We have a violation of this rule in the FollowArtistController.
- Each controller communicates with the client via a dedicated DTO with the following naming convention: 
Client -> Server = ...Request
Server -> Client = ...Response
So, for line 46 to 63, it could be an ArtistDetailsResponse.

src/main/resources/templates/frontend/artist-details.html
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE html>
+<html th:replace="~{frontend/fragments/layout :: layout (~{::body})}" lang="en" xmlns:th="http://www.w3.org/1999/xhtml">
+    <head th:include="admin/fragments/layout :: headerFragment">
+        <title id="pageTitle">Artist details | Metal Release Radar</title>
+    </head>
+    <body>
+        <div class="mb-3 mt-5">
```
Waginator: the headline jumps if you switch the page, for example from an artist details page to the profile page.

Waginator:
Currently, you have merge conflicts in your remote branch.

PR #71, opened 2019-11-25
Add Dockerfile and docker-compose to start all peripheral systems

Waginator:
With this PR a first Dockerfile for the Metal Release Radar is added. We'll need the Dockerfile later to build a Docker Image from the Spring Boot application. This should work right now, but I haven't tested it yet, because it will be built by our build pipeline later. 

Furthermore the application now has a docker-compose.yml. This allows us to start all required peripheral systems of the application on the developer machine without having to install a local database server. Also an instance of the Release Butler is started.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,11 @@
 !**/src/main/**
 !**/src/test/**
 
-# ignore all stuff in private folder
+# ignore all stuff in .private folder
 .private/
+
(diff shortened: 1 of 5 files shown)


Waginator:
[Create Dockerfile and docker-compose.yml](https://trello.com/c/T6zPNCze/32-create-dockerfile-and-docker-composeyml)

PR #73, opened 2019-11-25
Use javas duration

Waginator:
Removal of HttpClientConfigConstants and ExpirationTime. I've used Javas Duration class instead.


Waginator:
[Removal of HttpClientConfigConstants and ExpirationTime](https://trello.com/c/kuV6jsnT/9-removal-of-httpclientconfigconstants-and-expirationtime)

src/main/java/com/metalr2/config/resttemplate/ApacheHttpClientConfig.java
```diff
@@ -76,9 +78,9 @@ public void run() {
         // only if connection pool is initialised
         if (pool != null) {
           pool.closeExpiredConnections();
-          pool.closeIdleConnections(IDLE_CONNECTION_WAIT_TIME, TimeUnit.MILLISECONDS);
+          pool.closeIdleConnections(10, MINUTES);
```
hedgehog1833: The change from 30 seconds to 10 minutes is on purpose?

src/main/java/com/metalr2/config/resttemplate/ApacheHttpClientConfig.java
```diff
@@ -95,9 +97,9 @@ public TaskScheduler taskScheduler() {
   @Bean
   public CloseableHttpClient httpClient() {
     RequestConfig requestConfig = RequestConfig.custom()
-            .setConnectTimeout(CONNECTION_TIMEOUT)
-            .setConnectionRequestTimeout(REQUEST_TIMEOUT)
-            .setSocketTimeout(SOCKET_TIMEOUT)
+            .setConnectTimeout((int) Duration.ofSeconds(20).toMillis()) // the time for waiting until a connection is established
+            .setConnectionRequestTimeout((int) Duration.ofSeconds(20).toMillis()) // the time for waiting for a connection from connection pool
+            .setSocketTimeout((int) Duration.ofSeconds(30).toMillis()) // the time for waiting for data
```
hedgehog1833: Here the times are also changed, but only slightly.

src/main/java/com/metalr2/config/resttemplate/ApacheHttpClientConfig.java
Waginator: Yipp. We don't need an idle connection cleanup every 30 seconds.

src/main/java/com/metalr2/config/resttemplate/ApacheHttpClientConfig.java
Waginator: Yeah. I think 20 (or 30) seconds timeout is more than enough.

PR #75, opened 2019-12-01
Integration tests without authentication

Waginator:


Waginator:
[Authentication in tests should not be necessary](https://trello.com/c/oVVKSiba/29-authentication-in-tests-should-not-be-necessary)

PR #76, opened 2019-12-01
Reduce code duplication in rest controller

Waginator:


Waginator:
[Reduce code duplication in REST Controller](https://trello.com/c/eoTJyoV8/33-reduce-code-duplication-in-rest-controller)

PR #77, opened 2019-12-01
check signature of token

Waginator:


Waginator:
[Check the signature of the tokens in registration verification and reset password process](https://trello.com/c/ajIHW1gV/27-check-the-signature-of-the-tokens-in-registration-verification-and-reset-password-process)

PR #80, opened 2019-12-09
Get for search controller

hedgehog1833:
ArtistDetailsRestController and SearchArtistRestController are now using GET instead of POST
https://trello.com/c/N8YKBBai


hedgehog1833:
[Use GET instead of POST in ArtistDetailsRestController and SearchArtistRestController](https://trello.com/c/N8YKBBai/41-use-get-instead-of-post-in-artistdetailsrestcontroller-and-searchartistrestcontroller)

PR #86, opened 2019-12-21
Artist details card design

hedgehog1833:
New version of the artist details page. Yes, the navigation (and the rest) is still ugly, but functional. The image gallery looks quite nice already I think. But from now on it should be mostly styling as the functions are good for now.
https://trello.com/c/C65jtjIY


hedgehog1833:
[Revision of artist detail page](https://trello.com/c/C65jtjIY/15-revision-of-artist-detail-page)

PR #91, opened 2020-01-04
Display all users in admin area using Datatables jQuery Plugin

DanielW1987:
Before the PR becomes so huge, I'll create it now.

Currently implemented is the display of all users in the admin area. All users are queried via Rest and are stored on the client side in a JavaScript array (Datatables plugin does that for us). The pagination, searching and sorting is then completely client-side. This can also be done on the server side - but it is only worth the effort when we are talking about a large amount of data (>1000).

Following this PR I will implement the creation of an administrator and the activation or reactivation of a user.


DanielW1987:
[CRU for users](https://trello.com/c/h9U3tWoB/65-cru-for-users)

src/test/java/com/metalr2/web/controller/rest/UserRestControllerIT.java
```diff
@@ -0,0 +1,164 @@
...
+      when(userService.getAllUsers()).thenReturn(Collections.emptyList());
+
+      // when
+      requestHandler.doGet(ContentType.JSON);
+
+      // then
+      verify(userService, times((1))).getAllUsers();
+    }
+
+  }
```
hedgehog1833: Why don't you put this into one test? I have seen this in the UserServiceTest too.

DanielW1987: To keep the tests clean and small. I like to separate the tests by type: one test tests the values, another test verifies the mock interaction.

PR #92, opened 2020-01-05
Reintroduce mvc controller

hedgehog1833:
MVC controller for Search and ArtistDetails now functioning. They forward the url parameters to the html page which calls the RestController per JS.


hedgehog1833:
[Reintroduce MVC Controller for Search and ArtistDetails](https://trello.com/c/2YxIcGYy/68-reintroduce-mvc-controller-for-search-and-artistdetails)

src/main/java/com/metalr2/service/artist/ArtistsService.java
```diff
@@ -19,7 +19,7 @@
   boolean isFollowed(long discogsId);
   List<FollowArtistDto> findFollowedArtistsPerUser(String publicUserId);
 
-  Optional<ArtistNameSearchResponse> searchDiscogsByName(String artistQueryString, int page, int size);
+  Optional<SearchResponse> searchDiscogsByName(String artistQueryString, int page, int size);
```
DanielW1987: 👍 for renaming

src/main/java/com/metalr2/service/artist/ArtistsServiceImpl.java
```diff
@@ -112,22 +112,22 @@ public boolean isFollowed(long discogsId) {
...
     int itemsPerPage = discogsPagination.getItemsPerPage();
 
     Set<Long> alreadyFollowedArtists = findFollowedArtistsPerUser(currentUserSupplier.get().getPublicId()).stream().map(FollowArtistDto::getArtistDiscogsId)
         .collect(Collectors.toSet());
 
-    List<ArtistNameSearchResponse.ArtistSearchResult> dtoArtistSearchResults = artistSearchResults.getResults().stream()
-        .map(artistSearchResult -> new ArtistNameSearchResponse.ArtistSearchResult(artistSearchResult.getThumb(), artistSearchResult.getId(),
-                                                                                   artistSearchResult.getTitle(), alreadyFollowedArtists.contains(artistSearchResult.getId())))
+    List<SearchResponse.SearchResult> dtoSearchResults = artistSearchResults.getResults().stream()
+        .map(artistSearchResult -> new SearchResponse.SearchResult(artistSearchResult.getThumb(), artistSearchResult.getId(),
```
DanielW1987: maybe a static import of `SearchResult` makes it more easier to read ;)

src/main/resources/templates/frontend/fragments/layout.html
```diff
@@ -60,7 +62,7 @@
                 <a class="nav-link dropdown-toggle" href="#" id="navbarUserDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                   <img src="#" th:src="@{/images/dummy-profile-picture.jpg}" alt="Profile photo" height="30" width="30" class="avatar border border-white" />
                 </a>
-                <div class="dropdown-menu dropdown-menu-right" aria-labelledby="navbarUserDropdown">
+                <div class="dropdown-menu" aria-labelledby="navbarUserDropdown">
```
DanielW1987: why did you remove that?

DanielW1987:
Please resolve conflicts :)

src/main/resources/templates/frontend/fragments/layout.html
hedgehog1833: Stupidity ;-)

hedgehog1833:
I (really) don't have any conflicts oO

src/main/java/com/metalr2/web/controller/mvc/ArtistsController.java
```diff
@@ -0,0 +1,33 @@
...
+public class ArtistsController {
+
+  private static final String DEFAULT_QUERY = "";
+  private static final String DEFAULT_PAGE  = "1";
+  private static final String DEFAULT_SIZE  = "10";
+
+  @GetMapping(path = Endpoints.Frontend.SEARCH)
+  public ModelAndView showSearch(@RequestParam(value = "query", defaultValue = DEFAULT_QUERY) String query,
+                                 @RequestParam(value = "page", defaultValue = DEFAULT_PAGE) int page,
+                                 @RequestParam(value = "size", defaultValue = DEFAULT_SIZE) int size) {
```
DanielW1987: why do we need the query parameter in our mvc controller? We run into the same validation problems again?
Do we really need this? Can JavaScript not read the parameters directly from the URL? Then we don"t need this workaround.

src/main/java/com/metalr2/web/controller/mvc/ArtistsController.java
```diff
@@ -0,0 +1,33 @@
...
+    return new ModelAndView(ViewNames.Frontend.SEARCH, "searchRequest", new SearchRequest(query,page,size));
+  }
+
+  @GetMapping(path = "/{discogsId}")
+  public ModelAndView showArtistDetails(@PathVariable long discogsId) {
+    return new ModelAndView(ViewNames.Frontend.ARTIST_DETAILS, "discogsId", discogsId);
+  }
```
DanielW1987: I would suggest to use String instead of Long to avoid transformation errors in case of URL manipulation.

Is it really necessary to send the id value back to the client again?

src/main/resources/static/js/common.js
```diff
@@ -12,3 +12,61 @@ function registerLogoutListener() {
...
+
+/**
+ * Tasks when validation or ajax fails
+ * @param message       The message to be displayed
+ * @param containerId   The container to toggle the loader at
+ */
+function validationOrAjaxFailed(message, containerId) {
+    createNoResultsMessage(message);
+    toggleLoader(containerId);
+}
```
DanielW1987: The idea of `common.js` was to bundle javascript functionality that we need on nearly every page.
This seems to be very search specific.

DanielW1987:
@hedgehog1833 
Please consider all comments, not all were made within my review...

src/main/resources/static/js/common.js
hedgehog1833: Yeah I was unsure about that too. This is all stuff that is used by at least to files. I will put this into a separate file for when we can control wich js-file to load for which page.

PR #97, opened 2020-01-18
Pagination can now cope with lots of pages

hedgehog1833:
I built a naive fix for our pagination problem. Did not include any dependencies. With the MVC being back it is now easy to write some 'static' js that just show some pages. I built in some placeholders and always the next and previous to the current page.


hedgehog1833:
[Pagination is not prepared for lots of pages](https://trello.com/c/62SJi9Si/52-pagination-is-not-prepared-for-lots-of-pages)

PR #99, opened 2020-01-18
MyArtists page implemented

hedgehog1833:
I build a page with the followed artists just like the search page.

### DIFF
--- a/src/main/java/com/metalr2/web/controller/rest/MyArtistsRestController.java
+++ b/src/main/java/com/metalr2/web/controller/rest/MyArtistsRestController.java
@@ -6,7 +6,6 @@ import com.metalr2.web.dto.ArtistDto;
 import com.metalr2.web.dto.response.MyArtistsResponse;
 import com.metalr2.web.dto.response.Pagination;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.web.PageableDefault;
 import org.springframework.http.MediaType;
@@ -29,8 +28,8 @@ public class MyArtistsRestController {
   }
 
   @GetMapping(produces = {MediaType.APPLICATION_XML_VALUE, MediaType.APPLICATION_JSON_VALUE})
-  public ResponseEntity<MyArtistsResponse> getMyArtists(@PageableDefault(page = 1) Pageable pageable) {
-    List<ArtistDto> artists = artistsService.findFollowedArtistsForCurrentUser(PageRequest.of(pageable.getPageNumber() - 1, pageable.getPageSize()));
+  public ResponseEntity<MyArtistsResponse> getMyArtists(@PageableDefault Pageable pageable) {
+    List<ArtistDto> artists = artistsService.findFollowedArtistsForCurrentUser(pageable);
     long totalArtists = artistsService.countFollowedArtistsForCurrentUser();
     Pagination pagination = new Pagination(totalArtists, pageable.getPageNumber(), pageable.getPageSize());
     MyArtistsResponse response = new MyArtistsResponse(artists, pagination);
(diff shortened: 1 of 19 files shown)


hedgehog1833:
[Page MyArtists](https://trello.com/c/TWR59rSP/35-page-myartists)

src/main/resources/static/js/search.js
```diff
@@ -81,22 +81,22 @@ function createResultCards(searchResponse){
...
 
         const artistDetailsElement = document.createElement('a');
         artistDetailsElement.href = "/artists/" + searchResult.id;
         artistDetailsElement.text = "Details for " + searchResult.artistName;
         cardBody.append(artistDetailsElement);
 
-        const breakElement = document.createElement('br');
-        cardBody.append(breakElement);
+        const breakElement2 = document.createElement('br');
+        cardBody.append(breakElement2);
```
DanielW1987: `cardBody.append(breakElement);` is not possible?

src/main/resources/templates/frontend/fragments/layout.html
```diff
@@ -48,7 +49,7 @@
             </form>
             <ul class="navbar-nav navbar-nav-right">
               <li class="nav-item">
-                <a class="nav-link" href="#" th:href="@{/my-artists}" >My Artists</a>
+                <a class="nav-link" href="#" th:href="@{/my-artists?page=1&size=10}" >My Artists</a>
```
DanielW1987: do we really need the `?page=1&size=10`? We have default values in controller I thought...

src/main/java/com/metalr2/web/controller/rest/MyArtistsRestController.java
```diff
@@ -29,7 +30,7 @@ public MyArtistsRestController(ArtistsService artistsService) {
 
   @GetMapping(produces = {MediaType.APPLICATION_XML_VALUE, MediaType.APPLICATION_JSON_VALUE})
   public ResponseEntity<MyArtistsResponse> getMyArtists(@PageableDefault Pageable pageable) {
-    List<ArtistDto> artists = artistsService.findFollowedArtistsForCurrentUser(pageable);
+    List<ArtistDto> artists = artistsService.findFollowedArtistsForCurrentUser(PageRequest.of(pageable.getPageNumber()-1, pageable.getPageSize()));
```
DanielW1987: what about the Pageable parameter in other controllers...?

Idea:
javascript do the mapping. So in URL we see page=1 but javascript request rest endpoint with page=0.
This looks very hacky. You can create a trello card if you see this like me.

src/main/resources/static/js/my-artists.js
```diff
@@ -0,0 +1,214 @@
...
+    if (myArtists.pagination.currentPage !== myArtists.pagination.totalPages)
+      createPageLinkMyArtists(myArtists, myArtists.pagination.totalPages, listElement);
+  }
+
+  // Next link
+  if (myArtists.pagination.currentPage < myArtists.pagination.totalPages)
+    createPreviousOrNextItemMyArtists(myArtists, listElement, false);
+
+  document.getElementById('myArtistsContainer').append(navElement);
+}
```
DanielW1987: Make sense to have a util function for this...it's a lot of duplicate code.
Trello card please.

src/main/resources/static/js/search.js
hedgehog1833: You cannot append the same element twice, but in this case .clone() works!

src/main/resources/templates/frontend/fragments/layout.html
hedgehog1833: That esalated a bit. I removed the page and size and suddenly nothing worked anymore. We had some issues with the @PageableDeafult (and also our repository) using page = 0 als default. Our page (the pagination and so on) all think page 1 is the first. The discogs api worked just by chance because is gives the same page for 0 and 1. I made some adjustments and re-request a review.

src/main/java/com/metalr2/web/controller/rest/MyArtistsRestController.java
hedgehog1833: I saw this comment too late. But just javascript doesn't really help. The pagination links also have to start with 1. Everything else is just stupid. Like I said in my previous comment, I think we should overrule the sping guys decision to start pagination with 0. Discogs also works this way. I'll submit another version and we'll discuss it.

hedgehog1833:
Okay after thinking twice, of course we can map the pagination links to the spring defaults and just show the "normal" uman-readable ones... I still don't like it... Let's talk about that tomorrow.

hedgehog1833:
I rewrote the pagination logic and documente everything on Trello.

PR #104, opened 2020-02-02
Quickwin transfer code to user service

hedgehog1833:
The ResetPasswordController now does not handle that much business logic anymore. The UserService now also checks if the token exists or is expired.


hedgehog1833:
[Transfer code from ResetPasswordController to UserService.](https://trello.com/c/5lk3ePh1/39-transfer-code-from-resetpasswordcontroller-to-userservice)

PR #105, opened 2020-02-02
User admin area

DanielW1987:
This PR includes the following:
- Administrators can now be created in the administration area
- For each user there is a detail view (in dialog)

Besides the code, please also take a look at the backend and check if everything feels right.

The following remains to be done:
- Edit users, that is, deactivate/activate users and change roles (https://trello.com/c/iEGmTlRI)
- login log (https://trello.com/c/VxTNVHQ6 | https://trello.com/c/HHW097cf)
- client-side validation (https://trello.com/c/PYArwUSx)
- Role-level security for rest endpoints (only administrators are allowed to request endpoints) (https://trello.com/c/yd2Xcrft)


DanielW1987:
[CRU for users](https://trello.com/c/h9U3tWoB/65-cru-for-users)

src/main/java/com/metalr2/service/mapper/UserMapper.java
```diff
@@ -26,13 +26,12 @@ public UserEntity mapToEntity(UserDto dto) {
...
+    userDto.setRole(determineRoleName(entity));
 
     return userDto;
   }
 
-  private String determinRoleName(UserEntity userEntity) {
+  private String determineRoleName(UserEntity userEntity) {
     if (userEntity.isAdministrator() || userEntity.isSuperUser()) {
       return "Administrator";
     }
```
hedgehog1833: Why does the role has to be a String in the DTO?

src/main/java/com/metalr2/model/exceptions/AppExceptionsHandler.java
```diff
@@ -21,10 +24,10 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(AppExceptionsHandler.class);
 
-  @ExceptionHandler(value = ValidationException.class)
-  public ResponseEntity<ErrorResponse> handleBadRequests(ValidationException exception, WebRequest webRequest) {
+  @ExceptionHandler(value = MethodArgumentNotValidException.class)
```
hedgehog1833: :+1:

src/main/resources/static/css/admin.css
```diff
@@ -99,15 +99,3 @@ div.content {
 /* Rules for using icons as white on a dark background. */
 .material-icons.md-light { color: rgba(255, 255, 255, 1); }
 .material-icons.md-light.md-inactive { color: rgba(255, 255, 255, 0.3); }
```
hedgehog1833: Please let the curser be a pointer for the table rows.

src/test/java/com/metalr2/web/controller/rest/ReleasesRestControllerIT.java
```diff
@@ -130,7 +130,7 @@ void bad_requests() {
     // then
     validatableResponse
         .contentType(ContentType.JSON)
-        .statusCode(HttpStatus.BAD_REQUEST.value());
+        .statusCode(HttpStatus.UNPROCESSABLE_ENTITY.value());
```
hedgehog1833: Why did you choose 409? Just curious as I thought this was a clear bad request, but reading some more infos about http responses makes this also sounds a bit fitting... Still I'd prefer bad request (not a must have for approval though).

src/main/java/com/metalr2/service/mapper/UserMapper.java
DanielW1987: what else? 

I wanted a clear display name so that nothing has to be transformed with javascript. 

And on the client side, I wanted to hide the difference between administrator and superuser, because I found that confusing. Therefore the roles should be revised (https://trello.com/c/er30yUTr)

src/main/resources/static/css/admin.css
DanielW1987: Good point

src/test/java/com/metalr2/web/controller/rest/ReleasesRestControllerIT.java
DanielW1987: I read something about it, too. I find 409 a little more talking. That was my intention to change that. But the status code doesn't seem to have caught on in general and Bad Request is best practice - so I'm reverting it.

src/main/resources/static/css/admin.css
DanielW1987: see https://github.com/DanielW1987/metal-release-radar/pull/105/commits/7659b23123bbdd9b617384f80ebc497ca2cd03d9

src/test/java/com/metalr2/web/controller/rest/ReleasesRestControllerIT.java
DanielW1987: see https://github.com/DanielW1987/metal-release-radar/pull/105/commits/71e42b81d35e1f6c1762b3de0cad8fbeeba59f30

PR #106, opened 2020-02-08
SuperUser Role removed

hedgehog1833:
The superuser role is now removed. I also did some smaller refactorings. RemoveRole now checks if the role to be is removed is the last role. If so an exception is thrown. I still don't like it perfectly well because it feels kind of weird in usage in some places.
Do you think it would be better if an admin user always has both roles?


src/main/java/com/metalr2/model/user/UserEntity.java
```diff
@@ -99,6 +110,10 @@ public void setUserRoles(Set<UserRole> newUserRoles) {
   }
 
   public boolean removeUserRole(UserRole userRole) {
+    if (userRoles.equals(Set.of(userRole))) {
+      throw new IllegalArgumentException("At least one user role must be set!");
+    }
```
DanielW1987: It's more an `IllegalStateException` than `IllegalArgumentException`

src/main/java/com/metalr2/service/mapper/UserMapper.java
```diff
@@ -32,7 +32,7 @@ public UserDto mapToDto(UserEntity entity) {
   }
 
   private String determineRoleName(UserEntity userEntity) {
-    if (userEntity.isAdministrator() || userEntity.isSuperUser()) {
+    if (userEntity.isAdministrator()) {
```
DanielW1987: To remove the whole method we can give the `UserRole` enum a method to return the display name.

hedgehog1833: I removed the method but I had to add another one to the entity to determine the user's "highest" role.

PR #108, opened 2020-02-18
Updating user accounts

hedgehog1833:
Updating a user's role and status is now possible via the user administration. I pulled the last checkbox from the card into a new one because it does not really have anything to do with this topic and I wanted to finally get this done here.

### DIFF
--- a/src/main/resources/static/js/admin/users.js
+++ b/src/main/resources/static/js/admin/users.js
@@ -82,13 +82,15 @@ function createAdministrator () {
     $.post({
         url: '/rest/v1/users',
         data: createAdministratorCreateRequest(),
-        type: 'POST',
+        type: 'PUT',
         headers: {
             'Accept': 'application/json',
             'Content-Type': 'application/json',
         },
         success: onCreateAdministratorSuccess,
-        error: onCreateAdministratorError
+        error: function (errorResponse) {
+            onCreateUpdateError(errorResponse, '#create-admin-user-validation-area')
+        }
     });
 }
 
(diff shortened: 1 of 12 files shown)


hedgehog1833:
[Update a user account](https://trello.com/c/iEGmTlRI/55-update-a-user-account)

src/main/resources/static/js/admin/users.js
```diff
@@ -176,37 +191,59 @@ function updateUser () {
  * Sends the update request to the server.
  */
 function sendUpdateUserRequest() {
-    // ToDo: https://trello.com/c/iEGmTlRI
+    $.post({
+        url: '/rest/v1/users',
+        data: createUpdateUserRequest(),
+        type: 'POST',
```
DanielW1987: PUT

src/main/resources/static/js/admin/users.js
```diff
@@ -117,11 +119,12 @@ function onCreateAdministratorSuccess(createResponse) {
...
 /**
  * Error callback for creating a new administrator.
- * @param errorResponse The json response
+ * @param errorResponse     The json response
+ * @param validationAreaId  ID of the area to display errors (create/update)
  */
-function onCreateAdministratorError(errorResponse) {
-    resetCreateAdminUserValidationArea();
-    const validationMessageArea = $('#create-admin-user-validation-area');
+function onCreateUpdateError(errorResponse, validationAreaId) {
```
DanielW1987: I don't like that kind of mix in the name. I'd rather have one method more:

- onCreateError (calls onError)
- onUpdateError (calls onError)
- onError

src/test/java/com/metalr2/service/user/UserServiceTest.java
```diff
@@ -377,11 +407,56 @@ void update_user_for_not_existing_user() {
...
+
+    when(currentUserSupplier.get()).thenReturn(user);
+    when(userRepository.findByPublicId(PUBLIC_ID)).thenReturn(Optional.of(user));
+
+    // when
+    Throwable throwable = catchThrowable(() -> userService.updateUser(PUBLIC_ID, userDtoForUpdate));
+
+    // then
+    verify(userRepository, times(1)).findByPublicId(PUBLIC_ID);
+    verify(currentUserSupplier, times(1)).get();
```
DanielW1987: you already have test the method invocation in other tests. The test should be as minimal as possible to test the exception.

src/test/java/com/metalr2/service/user/UserServiceTest.java
```diff
@@ -377,11 +407,56 @@ void update_user_for_not_existing_user() {
...
```
DanielW1987: you already have test the method invocation in other tests. The test should be as minimal as possible to test the exception.

PR #112, opened 2020-03-07
Dependencies updated

hedgehog1833:
All dependencies updated. The only one which did not get the newest one (but was still updated to the current major) is Mockito because something seems to have changed there and I could not resolve it just now.


hedgehog1833:
[[MetalDetector] Update Versions of all Dependencies](https://trello.com/c/EXNjoSAY/83-metaldetector-update-versions-of-all-dependencies)

PR #116, opened 2020-03-14
Role level security for users endpoint

hedgehog1833:
The rest endpoint for users is now only accessible for administrators. Also the IntegrationTest for this endpoint was reduced to checking if security works. A unit test for controller functionality is introduced.


hedgehog1833:
[Role-level security for rest endpoints in UserRestController](https://trello.com/c/yd2Xcrft/90-role-level-security-for-rest-endpoints-in-userrestcontroller)

hedgehog1833:
I just pushed one more commit where I found out how to easily deny access to anonymous users on rest endpoints (instead of forwarding them to login). I also added a test for that.

hedgehog1833:
I found an easy way to deactivate the redirection for anonymous users on rest endpoints and added another test for it.

hedgehog1833:
[Deactivate redirect for unauthorized users for rest endpoints](https://trello.com/c/7ml74Bfa/120-deactivate-redirect-for-unauthorized-users-for-rest-endpoints)

PR #117, opened 2020-03-14
Releases table in admin area

hedgehog1833:
The releases table is now located in the admin area. The checkbox and date filters are removed. I added a details dialog with the info from the table but did not yet add more fields. The Import Button and the marking of duplicates are also not in the scope of this PR. I'll create separate cards for them. The ReleasesController was deleted as not needed anymore.


hedgehog1833:
[[MetalDetector] Releases in admin area](https://trello.com/c/2kf1ZyQJ/107-metaldetector-releases-in-admin-area)

src/main/resources/templates/admin/fragments/layout.html
```diff
@@ -26,6 +26,7 @@
     <script th:src="@{/webjars/datatables/js/jquery.dataTables.min.js}" src="#"></script>
     <script th:src="@{/webjars/datatables/js/dataTables.bootstrap4.min.js}" src="#"></script>
     <link th:href="@{/webjars/datatables/css/dataTables.bootstrap4.min.css}" href="#" rel="stylesheet" type="text/css" />
+    <script th:src="@{/js/releases.js}" src="#"></script>
```
DanielW1987: we use the same JavaScript file for frontend and admin area at the moment?

src/main/resources/templates/admin/releases/update.html
```diff
@@ -0,0 +1,87 @@
...
+                                </div>
+                            </div>
+
+                            <!-- Status -->
+                            <div class="form-group row">
+                                <label for="duplicate-checkbox" class="col-sm-4 col-form-label">Duplicate</label>
+                                <div class="col-sm-8">
+                                    <label for="duplicate-checkbox" hidden></label><input id="duplicate-checkbox" type="checkbox" name="duplicate"> is duplicate<br>
+                                </div>
+                            </div>
```
DanielW1987: I think we need a combo box instead of a checkbox. The state can be more than a simple check if it's a duplicate or not...

src/test/java/rocks/metaldetector/web/controller/mvc/ReleasesControllerIT.java
```diff
@@ -1,28 +0,0 @@
...
-  @Test
-  @DisplayName("Requesting '" + Endpoints.Frontend.RELEASES + "' should return the releases view")
-  @WithMockUser
-  void get_search_should_return_releases_view() throws Exception {
-    mockMvc.perform(get(Endpoints.Frontend.RELEASES))
-        .andExpect(view().name(ViewNames.Frontend.RELEASES))
-        .andExpect(model().size(0))
-        .andExpect(content().contentType("text/html;charset=UTF-8"));
-  }
-}
```
DanielW1987: I think we still need a test that checks if we get the correct page for a certain URL. Can you transform this into a unit test?

src/main/java/rocks/metaldetector/config/mvc/WebConfig.java
```diff
@@ -36,6 +37,7 @@ public void addViewControllers(ViewControllerRegistry registry) {
     registry.addViewController(Endpoints.AdminArea.IMPORT).setViewName(ViewNames.AdminArea.IMPORT);
     registry.addViewController(Endpoints.AdminArea.ANALYTICS).setViewName(ViewNames.AdminArea.ANALYTICS);
     registry.addViewController(Endpoints.AdminArea.SETTINGS).setViewName(ViewNames.AdminArea.SETTINGS);
+    registry.addViewController(Endpoints.AdminArea.RELEASES).setViewName(ViewNames.AdminArea.RELEASES);
```
DanielW1987: Wwe actually only use this mechanism for pages that we have not yet actively worked on. With the `USERS` I must have forgotten that :)

src/test/java/rocks/metaldetector/web/controller/mvc/ReleasesControllerIT.java
hedgehog1833: But the Controller is gone.... what's there to check? That the WebConfig works correctly? We did not do that for other pages.

src/main/java/rocks/metaldetector/config/mvc/WebConfig.java
hedgehog1833: Isn't the Releases page in this state now again? I deleted the mvc controller as it only returned the "empty" page anyway. So what should happen if the releases endpoint is called if it is not configured here?

src/main/resources/templates/admin/releases/update.html
hedgehog1833: Do you have any concrete in mind? Otherwise I'll just use 'OK' and 'Duplicate' for now

DanielW1987: Not set, Ok, Duplicate

Enough for now

src/main/java/rocks/metaldetector/config/mvc/WebConfig.java
DanielW1987: We use controller for each Page we have worked on...Maybe there's been a misunderstanding?

We now have a releases Page in admin area with useful content?

hedgehog1833:
Like we talked about: I reintroduced a controller for the releases page and a unit test for it.

PR #118, opened 2020-03-15
Refactor ArtistsService

hedgehog1833:
The ArtistService returns a DTO and the ArtistsRestController converts it to a Response. It still annoys me that the Service cannot just return a list, because we don't have the number of items in total and would then have to request Discogs a second time. But I guess this way is still nicer than before.


hedgehog1833:
[Refactor ArtistsService](https://trello.com/c/Oo4I8AkR/80-refactor-artistsservice)

DanielW1987:
I think you can close this PR and remove the branch. The problem will be solved with the new modules.

PR #123, opened 2020-03-28
Pagination repaired

hedgehog1833:
The error was in the DiscogsArtistSearchResultContainerTransformer. The current page has to be decremented here. Looks like you just forgot that when writing the Transformer. And the js and Controller values had to be turned to their old values.


hedgehog1833:
I think it is working as intended now, as it was before. The `DiscogsArtistSearchResultContainerTransformer` shifts once for discogs in the backend. This is the only thing we do discogs-specific. The indxe + 1 in the frontend is normal, as Spring and the backend work with the first page 0 and for the user we want to show this as '1'.
So from my point of view this is the GEFN state that we agreed on, until we have another frontend (or something else gets in the way)

PR #125, opened 2020-03-29
Error handling on client for user updates

hedgehog1833:
The error handling for failing user updates for the client is now working. The server answers with a bad request and the client shows the error message.

### DIFF
--- a/webapp/src/main/java/rocks/metaldetector/service/exceptions/AppExceptionsHandler.java
+++ b/webapp/src/main/java/rocks/metaldetector/service/exceptions/AppExceptionsHandler.java
@@ -37,6 +37,12 @@ public class AppExceptionsHandler {
     return new ResponseEntity<>(createErrorResponse(exception), new HttpHeaders(), HttpStatus.CONFLICT);
   }
 
+  @ExceptionHandler(value = IllegalArgumentException.class)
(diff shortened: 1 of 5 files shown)


webapp/src/main/java/rocks/metaldetector/service/exceptions/AppExceptionsHandler.java
```diff
@@ -37,6 +37,12 @@
     return new ResponseEntity<>(createErrorResponse(exception), new HttpHeaders(), HttpStatus.CONFLICT);
   }
 
+  @ExceptionHandler(value = IllegalArgumentException.class)
+  public ResponseEntity<ErrorResponse> handleIllegalArgumentException(IllegalArgumentException exception, WebRequest webRequest) {
+    log.warn(webRequest.getContextPath() + ": " + exception.getMessage());
+    return new ResponseEntity<>(createErrorResponse(exception), new HttpHeaders(), HttpStatus.CONFLICT);
+  }
+
```
DanielW1987: IllegalArgumentException is mapped to 400 Bad Request with a PR to come from me. 

I would suggest to create a new exception (`IllegalUserActionException`) and throw this exception instead of the `IllegalArgumentException` in `UserServiceImpl`. 

You can also use the same handle method for `UserAlreadyExistsException` and `IllegalUserActionException`.

hedgehog1833:
[Finalize error handling (update user)](https://trello.com/c/qAAayCr1/99-finalize-error-handling-update-user)

PR #127, opened 2020-03-29
CreatedBy and lastModifiedBy are now correctly set

hedgehog1833:
Seemed to be an issue during initialization, see here for more info: https://stackoverflow.com/questions/52907096/spring-boot-data-jpa-createdby-and-updatedby-not-populating-with-authenticatin
Works correctly now.

### DIFF
--- a/persistence/src/main/java/rocks/metaldetector/persistence/config/PersistenceModuleConfiguration.java
+++ b/persistence/src/main/java/rocks/metaldetector/persistence/config/PersistenceModuleConfiguration.java
@@ -15,25 +15,27 @@ import java.util.Optional;
 
 @Configuration
 @EnableJpaAuditing
-@EnableJpaRepositories(basePackages ={"rocks.metaldetector.persistence"})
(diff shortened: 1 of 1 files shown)


hedgehog1833:
[createdBy and lastModifiedBy is empty after creating a new user](https://trello.com/c/fjr6dS1L/89-createdby-and-lastmodifiedby-is-empty-after-creating-a-new-user)

hedgehog1833:
There is, I just checked it. Before created_by and last_modified_by were always ANONYMOUS. After the change the correct user is set there. Works also on update now.
The issue is the following:
On startup the Bean is created. The old code checked *on startup* for the current user which is not set yet (or the starting thread cannot access it or whatever, check the link 😉). It therefore gets to the last else and *always* returns a lambda that gives an Optional with ANONYMOUS.
The new code returns a lambda that *contains* the check for the username. So the check is run when the Bean is called instead of the Bean's creation. And then the user is known.

PR #128, opened 2020-03-30
Butler import button

hedgehog1833:
We can now import releases into the Butler via a button above the releases table. The commits are pretty straight forward: ButlerClient extended, ReleasesService extended, ReleasesRestController extended and then frontend finished.


hedgehog1833:
[Import Button to trigger Butler import](https://trello.com/c/VUMMePvm/123-import-button-to-trigger-butler-import)

webapp/src/main/resources/static/js/admin/users.js
```diff
@@ -256,6 +256,6 @@ function resetUpdateUserForm() {
  */
 function resetValidationArea(validationAreaId) {
     const validationMessageArea = $(validationAreaId);
-    validationMessageArea.removeClass('alert alert-danger');
+    validationMessageArea.removeClass('alert alert-danger alert-success');
```
DanielW1987, line 259: Sure we need that here? Why?

webapp/src/main/java/rocks/metaldetector/web/controller/rest/ReleasesRestController.java
```diff
@@ -32,4 +37,12 @@
     List<DetectorReleasesResponse> response = releasesResponseTransformer.transformListOf(releaseDtos);
     return ResponseEntity.ok(response);
   }
+
+  @GetMapping(path = "/import",
```
DanielW1987, line 41: The idea behind `Endpoints.Rest` is that we have a central location where we see all of our defined endpoints. 

So this class here needs two constants in `Endpoints.Rest`:
- `Endpoints.Rest.QUERY_RELEASES`
- `Endpoints.Rest.IMPORT_RELEASES`

webapp/src/test/java/rocks/metaldetector/web/controller/rest/ReleasesRestControllerTest.java
```diff
@@ -0,0 +1,211 @@
...
+    validatableResponse
+        .contentType(ContentType.JSON)
+        .statusCode(HttpStatus.SERVICE_UNAVAILABLE.value());
+  }
+
+  @Test
+  @DisplayName("Should call release service on import")
+  void should_call_release_service() {
+    // when
+    restAssuredUtils.doGet("/import");
```
DanielW1987, line 158: don't forget to include the new constant here for all `/import` occurrences

webapp/src/test/java/rocks/metaldetector/web/controller/rest/ReleasesRestControllerTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+    // when
+    var validatableResponse = restAssuredUtils.doGet("/import");
+
+    // then
+    validatableResponse
+        .contentType(ContentType.JSON)
+        .statusCode(HttpStatus.SERVICE_UNAVAILABLE.value());
+  }
+}
```
DanielW1987, line 211: This is more a test of the ExceptionHandler. To test the ExceptionHandler, you should build a simple RestController, where certain exceptions are triggered. Afterwards you can check if they are mapped to the correct status code. 

Why not do that here?

1) It does not belong to the unit under test
2) If we need to refactor the exception handler we have to refactor several test classes

I'll make an extra card for it

webapp/src/main/resources/static/js/admin/users.js
hedgehog1833: The releases.js also uses this function. I moved it to util.js so it is clearer that this function resets different validation areas.

PR #129, opened 2020-04-01
All release details available

hedgehog1833:
All infos saved in the butler DB can now be seen in the backend releases table.


hedgehog1833:
[More details for releases](https://trello.com/c/KUHCZrH8/122-more-details-for-releases)

butler/src/main/java/rocks/metaldetector/butler/api/ButlerReleaseEntityRecordState.java
```diff
@@ -0,0 +1,20 @@
...
+  private final String name;
+
+  ButlerReleaseEntityRecordState(String name) {
+    this.name = name;
+  }
+
+  public String toDisplayString() {
+    return this.name;
+  }
+}
```
DanielW1987, line 20: I would actually like to get rid of the enum. We will always have a breaking change in the Detector when the butler suddenly sends a new value for the enum. Isn't the string enough for us here? We only use it for the display.

butler/src/main/java/rocks/metaldetector/butler/api/ButlerReleaseSource.java
```diff
@@ -0,0 +1,17 @@
...
+  private final String name;
+
+  ButlerReleaseSource(String name) {
+    this.name = name;
+  }
+
+  public String toDisplayString() {
+    return this.name;
+  }
+}
```
DanielW1987, line 17: I would actually like to get rid of the enum. We will always have a breaking change in the Detector when the butler suddenly sends a new value for the enum. Isn't the string enough for us here? We only use it for the display.

butler/src/main/java/rocks/metaldetector/butler/api/ButlerReleaseType.java
```diff
@@ -0,0 +1,24 @@
...
+  private final String name;
+
+  ButlerReleaseType(String name) {
+    this.name = name;
+  }
+
+  public String toDisplayString() {
+    return this.name;
+  }
+}
```
DanielW1987, line 24: I would actually like to get rid of the enum. We will always have a breaking change in the Detector when the butler suddenly sends a new value for the enum. Isn't the string enough for us here? We only use it for the display.

butler/src/main/java/rocks/metaldetector/butler/client/transformer/ButlerReleaseResponseTransformer.java
```diff
@@ -22,6 +25,24 @@ private ReleaseDto transformRelease(ButlerRelease release) {
...
+    return type != null ? type.toDisplayString() : null;
+  }
+
+  private String transformReleaseSource(ButlerReleaseSource source) {
+    return source != null ? source.toDisplayString() : null;
+  }
+
+  private String transformReleaseState(ButlerReleaseEntityRecordState state) {
+    return state != null ? state.toDisplayString() : null;
+  }
```
DanielW1987, line 47: If we kick out the Enums, we have to figure out how to get great display names here. 

Option 1) 
Not at all. We do it on the front end.

Option 2) 
We use Apache Commons Text for that: <https://commons.apache.org/proper/commons-text/javadocs/api-1.2/org/apache/commons/text/WordUtils.html#capitalizeFully-java.lang.String->

hedgehog1833: I went the apache way with this.

butler/src/main/java/rocks/metaldetector/butler/client/transformer/ButlerReleaseResponseTransformer.java
```diff
@@ -26,23 +26,15 @@ private ReleaseDto transformRelease(ButlerRelease release) {
...
             .estimatedReleaseDate(release.getEstimatedReleaseDate())
             .genre(release.getGenre())
-            .type(transformReleaseType(release.getType()))
+            .type(prettyPrintValue(release.getType()))
             .metalArchivesArtistUrl(release.getMetalArchivesArtistUrl())
             .metalArchivesAlbumUrl(release.getMetalArchivesAlbumUrl())
-            .source(transformReleaseSource(release.getSource()))
-            .state(transformReleaseState(release.getState()))
+            .source(prettyPrintValue(release.getSource()))
+            .state(prettyPrintValue(STATE_NOT_SET))
```
DanielW1987, line 33: why not `release.getState()`?

hedgehog1833: Then the state was empty in the front-end. And as long as we don't do anything with it I just set it here.

PR #131, opened 2020-04-02
Unify Configuration between Butler and Detector

DanielW1987:
I moved the whole RestTemplate configuration (we're using the Apache HTTP Client below) to the support module. 

Discogs now has its own RestTemplate with its own credentials interceptor.


DanielW1987:
[Unify Configuration between Butler and Detector](https://trello.com/c/COJY57AW/129-unify-configuration-between-butler-and-detector)

PR #132, opened 2020-04-04
Email notifications for new releases

hedgehog1833:
I added a NotificationService that can sends NewReleasesMails to all or a certain user. The notifyAll() method is scheduled to run every Sunday at 4 a.m. but can also be triggered via a NotificationRestController.
I also add a few tests where non existed but testing the logging if mail sending fails is annoying. Also I'm not entirely happy with the Exception Handling in the JavaMailEmailService: For example if creating the MimeMessage fails we still try to send the mail. Shouldn't we react some more to these exceptions (on creating and sending)? The Service currently doesn't even notice that.
Also: do we need another integration test for the NotificationController's endpoints? They too can only be called by admins.


hedgehog1833:
[Email notification](https://trello.com/c/sfYu9y7k/111-email-notification)

webapp/src/main/java/rocks/metaldetector/web/controller/rest/NotificationRestController.java
```diff
@@ -0,0 +1,30 @@
...
+  public ResponseEntity<Void> notifyUser(@PathVariable String publicUserId) {
+    notificationService.notifyUser(publicUserId);
+    return ResponseEntity.ok().build();
+  }
+
+  @GetMapping(path = Endpoints.Rest.NOTIFY)
+  public ResponseEntity<Void> notifyAllUsers() {
+    notificationService.notifyAllUsers();
+    return ResponseEntity.ok().build();
+  }
```
DanielW1987, line 29: Should be `POST` because it's not only a query...

webapp/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
DanielW1987, line 1: A comment on what exactly this is needed for would be good.

webapp/src/test/java/rocks/metaldetector/service/email/ConsoleEmailServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+  @DisplayName("Sending an email to Console should interact with MailConfig as expected")
+  void send_email_should_use_mail_config() {
+    // given
+    AbstractEmail email = new NewReleasesEmail("john.doe@example.com", "JohnDoe", Collections.emptyList());
+
+    // when
+    emailService.sendEmail(email);
+
+    // then
+    String result = verify(mailConfig, times(1)).getFromEmail();
```
DanielW1987, line 81:
```suggestion
    verify(mailConfig, times(1)).getFromEmail();
```

webapp/src/test/java/rocks/metaldetector/web/controller/rest/NotificationRestControllerTest.java
```diff
@@ -0,0 +1,91 @@
...
+@ExtendWith(MockitoExtension.class)
+class NotificationRestControllerTest implements WithAssertions {
+
+  @Mock
+  NotificationService notificationService;
+
+  @InjectMocks
+  NotificationRestController underTest;
+
+  RestAssuredMockMvcUtils restAssuredUtils;
```
DanielW1987, line 32: Unfortunately, we have no private by default in Java yet ;)

DanielW1987:
FollowArtistDto was deleted, why?

webapp/src/test/java/rocks/metaldetector/service/email/ConsoleEmailServiceTest.java
hedgehog1833: That gives a warning, that the result of `getFromEmail()` is ignored. We have this a few times in our tests. In my opinion we could also ommit this test, as it just checks if the config file is used...

hedgehog1833:
The FollowArtistDto was unused.

PR #133, opened 2020-04-12
Preparations for Heroku Deployment

DanielW1987:
- Configuration for application setup simplified
- Conversion to PostgreSQL
- Conversion to application.yml
- 3 profiles: Default (the old `dev`), Preview, Prod
- enforce https redirection
- Caching Dependencies on the Classpath (useful for later use cases)


DanielW1987:
[[Detector] Heroku Deployment - Preview Stage](https://trello.com/c/bGaMkpAf/128-detector-heroku-deployment-preview-stage)

DanielW1987:
We would have needed the caching if we had included request throttling here as well. But with the Detector this makes no sense, because we have a lot of requests (css, js, images, rest calls). So I removed the request throttling again. We can certainly use the cache mechanism for future use cases - that's why it's still in there.

PR #134, opened 2020-04-13
Disabling account registration for preview and prod profile

DanielW1987:


DanielW1987:
[Deactivate registration temporarily](https://trello.com/c/C0cWEZ7M/148-deactivate-registration-temporarily)

webapp/src/test/java/rocks/metaldetector/web/controller/mvc/authentication/PreviewRegistrationControllerTest.java
```diff
@@ -0,0 +1,46 @@
...
+  private PreviewRegistrationController underTest = new PreviewRegistrationController();
+  private RestAssuredMockMvcUtils restAssuredUtils;
+
+  @BeforeEach
+  void setup() {
+    restAssuredUtils = new RestAssuredMockMvcUtils(Endpoints.Guest.REGISTER);
+    RestAssuredMockMvc.standaloneSetup(
+            underTest,
+            springSecurity((request, response, chain) -> chain.doFilter(request, response)),
+            exceptionResolver()
```
hedgehog1833, line 31: The exception resolver is not needed when not tests with exception exist. Please also remove the interface above.

webapp/src/main/resources/templates/guest/auth/disabled-register.html
```diff
@@ -0,0 +1,43 @@
...
+      </div>
+
+      <!-- Verify Password -->
+      <div class="form-label-group">
+        <input type="password" id="verifyPlainPassword" class="form-control" name="verifyPlainPassword" disabled placeholder="" autocomplete="new-password">
+        <label for="verifyPlainPassword">Verify Password</label>
+      </div>
+
+      <!-- Register -->
+      <button class="btn btn-lg btn-dark btn-block mt-4" type="submit" disabled>Join Metal Detector</button>
```
hedgehog1833, line 36: I would just leave this out. Why should we have the fields if noboby can register? Just the text and the already registered stuff.

webapp/src/test/java/rocks/metaldetector/web/controller/mvc/authentication/PreviewRegistrationControllerTest.java
DanielW1987: The `SecurityMockMvcConfigurer` is needed.

webapp/src/main/resources/templates/guest/auth/disabled-register.html
DanielW1987: Only the narrow text does not look good. I had looked at that first and then decided on the variants with the deactivated fields.

PR #135, opened 2020-04-13
Butler JWT support

hedgehog1833:
The release butler module now has it's own RestTemplace and interceptor and sends a token that is provided as an environment variable. The access tokens (just the value) are removed from logging.


hedgehog1833:
[Configure Detector to send token](https://trello.com/c/omx3kOh3/144-configure-detector-to-send-token)

PR #137, opened 2020-04-15
Integration tests to unit tests

hedgehog1833:
These are all of the missing integration tests converted to unit tests except for the LoginController. I guess we have to keep something on the integration test level there, but I'll do that in a separate PR. These are all quite straight forward here.


hedgehog1833:
[IntegrationTests to UnitTests](https://trello.com/c/x6hnzTvO/121-integrationtests-to-unittests)

PR #139, opened 2020-04-19
Login logging

hedgehog1833:
The last successful login is logged. The last five last failed logins are logged. Users are disabled after five failed login attempts.
My first attempt was to put this into the AuthenticationSuccess/FailedHandlers but for obvious reasons on failure you don't have the logged in user. So now i react to Springs events.
I also tried to give the UserServiceTest some more structure but I was too lazy to refactor all the tests there (making the single test smaller).


webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
```diff
@@ -203,6 +203,31 @@ public void changePassword(String tokenString, String newPassword) {
...
+    userEntity.setLastLogin(LocalDateTime.now());
+    userEntity.clearFailedLogins();
+
+    userRepository.save(userEntity);
+  }
+
+  @Override
+  public void handleFailedLogin(String username) {
+    UserEntity userEntity = userRepository.findByUsername(username)
+        .orElseThrow(() -> new ResourceNotFoundException(UserErrorMessages.USER_WITH_ID_NOT_FOUND.toDisplayString()));
```
DanielW1987, line 220: The exception cannot stay here, because there is no place to handle this exception. This simply interrupts the program flow here and the user no longer sees in the frontend that he has entered an incorrect user name.

We already have the handling of an incorrectly entered user name in another place. So here a simple if is enough.

webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
```diff
@@ -203,6 +203,31 @@ public void changePassword(String tokenString, String newPassword) {
...
+
+    userEntity.setLastLogin(LocalDateTime.now());
+    userEntity.clearFailedLogins();
+
+    userRepository.save(userEntity);
+  }
+
+  @Override
+  public void handleFailedLogin(String username) {
+    UserEntity userEntity = userRepository.findByUsername(username)
```
DanielW1987, line 219: And what if I want to authenticate with my email? Have a look at method `findByEmailOrUsername(String emailOrUsername)` in `UserService`.

PR #144, opened 2020-04-22
Increase jQuery version

DanielW1987:
I looked at the generated detector jar once and there were two jquery jars in different versions. With the latest version there is no conflict anymore. Let's see if this fixes the problem.


DanielW1987:
[[Detector] Could not find any content for '/webjars/jquery/jquery.min.js'](https://trello.com/c/vEVVU7jV/154-detector-could-not-find-any-content-for-webjars-jquery-jqueryminjs)

PR #145, opened 2020-04-30
use update instead of validate for ddl-auto

DanielW1987:
With the `update` we will have to test in the future whether we do not destroy the database when we make changes to the persistence layer.


DanielW1987:
[[Detector] Create database with sql script](https://trello.com/c/upm8HRSH/155-detector-create-database-with-sql-script)

PR #146, opened 2020-05-01
set mail from header

DanielW1987:
Our e-mails are currently still blocked by GMail, for example, because we do not send the From header.


DanielW1987:
[Sending Emails with Gmail doesn't work](https://trello.com/c/v3KKBJ2G/147-sending-emails-with-gmail-doesnt-work)

PR #149, opened 2020-05-02
Protect login from brute force

hedgehog1833:
Failed logins are cached for 24 hours (hashed IP and the amount of attemps). If there are more than five failed attempts the IP is blocked for 24 hours.

### DIFF
--- a/webapp/src/main/java/rocks/metaldetector/security/LoginAttemptService.java
+++ b/webapp/src/main/java/rocks/metaldetector/security/LoginAttemptService.java
@@ -15,7 +15,7 @@ public class LoginAttemptService {
 
   private final CacheManager cacheManager;
 
-  @CacheEvict(value = "failedLogins")
+  @CacheEvict(value = FAILED_LOGINS_CACHE)
   public void loginSucceeded(String ipHash) {
   }
 
--- a/webapp/src/main/java/rocks/metaldetector/security/LoginAttemptService.java
+++ b/webapp/src/main/java/rocks/metaldetector/security/LoginAttemptService.java
@@ -3,6 +3,7 @@ package rocks.metaldetector.security;
 import lombok.AllArgsConstructor;
 import org.springframework.cache.Cache;
 import org.springframework.cache.CacheManager;
+import org.springframework.cache.annotation.CacheEvict;
 import org.springframework.stereotype.Service;
 
 @Service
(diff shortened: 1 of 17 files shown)


hedgehog1833:
[[Detector] Captcha after too many failed logins](https://trello.com/c/9NdqWhND/153-detector-captcha-after-too-many-failed-logins)

webapp/src/test/java/rocks/metaldetector/security/LoginAttemptServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.springframework.cache.CacheManager;
+
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static rocks.metaldetector.security.LoginAttemptService.FAILED_LOGINS_CACHE;
+
+@ExtendWith(MockitoExtension.class)
+class LoginAttemptServiceTest implements WithAssertions {
```
DanielW1987, line 22: Please have a look at the namings in this class. In my opinion, no user or user account is blocked. It's only the IP.

webapp/src/test/java/rocks/metaldetector/service/user/UserServiceTest.java
```diff
@@ -345,6 +358,21 @@ void get_user_by_email_or_username_for_not_existing_user() {
       assertThat(userDto).isEmpty();
     }
 
+    @Test
+    @DisplayName("Requesting a blocked user by email or username should throw an exception")
+    void get_user_by_email_or_username_for_blocked_user() {
```
DanielW1987, line 363:
```suggestion
    @DisplayName("Requesting a user by email or username via blocked IP should throw an exception")
    void get_user_by_email_or_username_from_blocked_ip() {
```

webapp/src/main/java/rocks/metaldetector/security/LoginAttemptService.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public boolean isBlocked(String ipHash) {
+    Cache cache = cacheManager.getCache(FAILED_LOGINS_CACHE);
+
+    if (cache != null) {
+      Cache.ValueWrapper valueWrapper = cache.get(ipHash);
+      return valueWrapper != null && (int) valueWrapper.get() >= MAX_ATTEMPTS;
+    }
+    return true;
```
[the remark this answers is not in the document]
hedgehog1833: I would say this is a `should not happen` anyway and from a security point of view I would rather block too much than not enough.

PR #150, opened 2020-05-11
Removing followed artist entity

hedgehog1833:
The FollowArtistEntity was removed and replaced by a many-to-many relationship between UserEntity and ArtistEntity. A FollowArtistService was implemented and some functionality taken from the ArtistService.


hedgehog1833:
[[Detector] Refactor FollowArtistEntity](https://trello.com/c/rDehy9tf/138-detector-refactor-followartistentity)

persistence/src/test/java/rocks/metaldetector/persistence/domain/user/UserEntityTest.java
```diff
@@ -202,6 +203,72 @@ void update_password_with_empty_value_should_throw_exception() {
...
+    @Test
+    @DisplayName("Following an artist add the artist to user and user to artist")
+    void adding_artist_adds_artist_and_user() {
+      // given
+      UserEntity user = UserFactory.createUser("user", "email");
+      ArtistEntity artist = ArtistFactory.withDiscogsId(1L);
+      user.addFollowedArtist(artist);
+
+      // when
+      user.addFollowedArtist(artist);
```
DanielW1987, line 219:
```suggestion

      // when
      user.addFollowedArtist(artist);
```

webapp/src/main/java/rocks/metaldetector/service/notification/NotificationServiceImpl.java
```diff
@@ -34,8 +32,7 @@ public void notifyAllUsers() {
   @Override
   public void notifyUser(String publicUserId) {
     UserDto user = userService.getUserByPublicId(publicUserId);
-    List<String> followedArtistsNames = artistsService.findFollowedArtistsPerUser(publicUserId)
-        .stream().map(ArtistDto::getArtistName).collect(Collectors.toList());
+    List<String> followedArtistsNames = user.getFollowedArtists().stream().map(ArtistDto::getArtistName).collect(Collectors.toList());
```
DanielW1987, line 35: Can you test if it still works? I think the lazy loading will make this throwing an exception. Better is to use the FollowArtistService and implement a method `findFollowedArtistsOfUser(String` publicUserId)` and access the repository directly so that the access to the followed artists is wrapped in one transaction.

webapp/src/main/java/rocks/metaldetector/service/user/UserDto.java
```diff
@@ -24,5 +26,6 @@
   private LocalDateTime createdDateTime;
   private LocalDateTime lastModifiedDateTime;
   private String lastModifiedBy;
+  private List<ArtistDto> followedArtists;
```
DanielW1987, line 29: Do we really need this here?

The response could be quite big if we query all users in the Admin Area, for example.

Possibly it makes sense to provide the followed artists via another DTO.

webapp/src/test/java/rocks/metaldetector/service/user/UserServiceTest.java
```diff
@@ -409,25 +425,38 @@ void get_all_active_users_uses_user_repository() {
...
+      UserDto userDto3 = UserDtoFactory.createUser("a2", ROLE_USER, true);
+      UserDto userDto4 = UserDtoFactory.createUser("c1", ROLE_ADMINISTRATOR, false);
+      UserDto userDto5 = UserDtoFactory.createUser("c2", ROLE_ADMINISTRATOR, true);
+      UserDto userDto6 = UserDtoFactory.createUser("a3", ROLE_ADMINISTRATOR, true);
+      when(userTransformer.transform(user1)).thenReturn(userDto1);
+      when(userTransformer.transform(user2)).thenReturn(userDto2);
+      when(userTransformer.transform(user3)).thenReturn(userDto3);
+      when(userTransformer.transform(user4)).thenReturn(userDto4);
+      when(userTransformer.transform(user5)).thenReturn(userDto5);
+      when(userTransformer.transform(user6)).thenReturn(userDto6);
```
DanielW1987, line 447: A spy can safe time ;)

webapp/src/main/java/rocks/metaldetector/service/user/UserDto.java
hedgehog1833: As I refactored the NotificationService anyway I could just throw this out.

webapp/src/main/java/rocks/metaldetector/service/notification/NotificationServiceImpl.java
hedgehog1833: Was too lazy to test the old state and just rewrote it. Fits with the change from your comment in the UserDto 👍

PR #151, opened 2020-05-12
Update to Java 14

hedgehog1833:
We are now using Java 14. Please check the pom.xml . For some I was not able to pull groovy 3 when using `groovy.version`, I had to use the number to get it work (see commented lines).


hedgehog1833:
[[Detector] Move to Java 14](https://trello.com/c/KWRNk3g0/140-detector-move-to-java-14)

NilsDammenhayn:
Your tip worked, I renamed `groovy.version` to `custom-groovy.version` and now the correct one is used.

PR #153, opened 2020-05-21
Don't use port for links in email if profile is !default

DanielW1987:


DanielW1987:
[Don't use port for links in email if profile is !default](https://trello.com/c/0aQlqQvs/166-dont-use-port-for-links-in-email-if-profile-is-default)

PR #157, opened 2020-06-20
Add thymeleaf dialect

DanielW1987:
We now use a Thymeleaf dialect. This makes the use of layouts a bit more intuitive and we can now include JS files only on specific pages.

It is not necessary to make a thorough review here. On all HTML pages I had to change the header section.

We now have a new dependency and groovy is now also on the classpath for the productive code (previously only for testing).


DanielW1987:
[Include JavaScript and CSS files only on the pages where they are actually needed.](https://trello.com/c/OS3QBFFS/69-include-javascript-and-css-files-only-on-the-pages-where-they-are-actually-needed)

PR #162, opened 2020-07-09
Fix redirection to home page

hedgehog1833:
Revisiting the application while logged in redirects to home page (instead of status page).


hedgehog1833:
[Fix Redirecring Behaviour](https://trello.com/c/ZkPicMje/180-fix-redirecring-behaviour)

PR #170, opened 2020-07-19
Refactoring of artist id

hedgehog1833:
In this PR the `externalId` for artists is introduced together with a source from where the artists are fetched.
The externalId replaces the discogsId and will also be used for Spotify. The source value is currently hardcoded to discogs. Setting its dynamically will be done in the spotify PR as there would be unnecessary conflicts otherwise.

### DIFF
--- a/discogs/src/main/java/rocks/metaldetector/discogs/client/DiscogsArtistSearchRestClientImpl.java
+++ b/discogs/src/main/java/rocks/metaldetector/discogs/client/DiscogsArtistSearchRestClientImpl.java
@@ -52,21 +52,21 @@ public class DiscogsArtistSearchRestClientImpl implements DiscogsArtistSearchRes
   }
 
   @Override
-  public DiscogsArtist searchById(long artistId) {
-    if (artistId <= 0) {
-      throw new IllegalArgumentException("artistId must not be negative!");
+  public DiscogsArtist searchById(String externalId) {
+    if (externalId.isBlank()) {
+      throw new IllegalArgumentException("externalId must not be blank!");
     }
 
     ResponseEntity<DiscogsArtist> responseEntity = discogsRestTemplate.getForEntity(
             discogsConfig.getRestBaseUrl() + ARTIST_ID_SEARCH_URL_FRAGMENT,
             DiscogsArtist.class,
-            artistId
+            externalId
     );
 
     DiscogsArtist discogsArtist = responseEntity.getBody();
     var shouldNotHappen = discogsArtist == null || !responseEntity.getStatusCode().is2xxSuccessful();
     if (shouldNotHappen) {
-      throw new ExternalServiceException("Could not get artist for artistId '" + artistId + "' (Response code: " + responseEntity.getStatusCode() + ")");
+      throw new ExternalServiceException("Could not get artist for externalId '" + externalId + "' (Response code: " + responseEntity.getStatusCode() + ")");
     }
 
     return discogsArtist;
--- a/discogs/src/main/java/rocks/metaldetector/discogs/client/DiscogsArtistSearchRestClientImpl.java
+++ b/discogs/src/main/java/rocks/metaldetector/discogs/client/DiscogsArtistSearchRestClientImpl.java
@@ -53,7 +53,7 @@ public class DiscogsArtistSearchRestClientImpl implements DiscogsArtistSearchRes
 
   @Override
   public DiscogsArtist searchById(String externalId) {
-    if (externalId.isEmpty()) {
+    if (externalId == null || externalId.isEmpty()) {
       throw new IllegalArgumentException("externalId must not be empty");
     }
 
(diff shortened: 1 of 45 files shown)


hedgehog1833:
[Spotify Search](https://trello.com/c/lIsyI6P8/189-spotify-search)

discogs/src/main/java/rocks/metaldetector/discogs/client/DiscogsArtistSearchRestClientImpl.java
```diff
@@ -52,21 +52,21 @@ public DiscogsArtistSearchResultContainer searchByName(String artistQueryString,
   }
 
   @Override
-  public DiscogsArtist searchById(long artistId) {
-    if (artistId <= 0) {
-      throw new IllegalArgumentException("artistId must not be negative!");
+  public DiscogsArtist searchById(String externalId) {
+    if (externalId.isEmpty()) {
```
DanielW1987, line 56:
```suggestion
    if (externalId == null || externalId.isEmpty()) {
```

webapp/src/test/resources/integrationtest.properties
```diff
@@ -13,4 +13,6 @@ security.remember-me-secret=test-remember-me-secret
 
 spring.mail.host=test-mail-host
 spring.mail.username=test-mail-username
-spring.mail.password=test-mail-password
\ No newline at end of file
+spring.mail.password=test-mail-password
+
+server.port=8080
```
DanielW1987, line 18: why do we need that?

webapp/src/test/java/rocks/metaldetector/web/controller/rest/ArtistsRestControllerTest.java
```diff
@@ -34,7 +35,8 @@
 @ExtendWith(MockitoExtension.class)
 class ArtistsRestControllerTest implements WithAssertions {
 
-  private static final long VALID_ARTIST_ID = 252211L;
+  private static final String VALID_EXTERNLA_ID = "252211";
```
DanielW1987, line 38:
```suggestion
  private static final String VALID_EXTERNAL_ID = "252211";
```

webapp/src/main/java/rocks/metaldetector/web/controller/rest/ArtistsRestController.java
```diff
@@ -32,15 +32,18 @@
...
+      throw new IllegalArgumentException("Artist source must be set");
+    }
+    followArtistService.follow(externalId, source);
     return ResponseEntity.ok().build();
   }
 
-  @PostMapping(path = Endpoints.Rest.UNFOLLOW + "/{discogsId}")
-  public ResponseEntity<Void> handleUnfollow(@PathVariable long discogsId) {
-    followArtistService.unfollow(discogsId);
+  @PostMapping(path = Endpoints.Rest.UNFOLLOW + "/{externalId}")
```
DanielW1987, line 44: Why we don't need the source for unfollowing?

webapp/src/main/java/rocks/metaldetector/web/controller/rest/ArtistsRestController.java
```diff
@@ -32,15 +32,18 @@
     return ResponseEntity.ok(result);
   }
 
-  @PostMapping(path = Endpoints.Rest.FOLLOW + "/{discogsId}")
-  public ResponseEntity<Void> handleFollow(@PathVariable long discogsId) {
-    followArtistService.follow(discogsId);
+  @PostMapping(path = Endpoints.Rest.FOLLOW + "/{externalId}")
```
DanielW1987, line 35: For me it's more a "/rest/v1/follow/discogs/12345"

webapp/src/main/java/rocks/metaldetector/web/controller/rest/ArtistsRestController.java
NilsDammenhayn: Because we only need the soure on saving the artist. When unfollowing we don't care where the artist came from.

webapp/src/main/java/rocks/metaldetector/web/controller/rest/ArtistsRestController.java
NilsDammenhayn: Hmmm yeah does look better. But with unfollowing still the "old way"?

webapp/src/test/resources/integrationtest.properties
NilsDammenhayn: The integration tests for preview and prod complained that the port is missing and it worked after I added it 🤷‍♂️ Could be related to the spring-boot update to 2.3?

webapp/src/test/java/rocks/metaldetector/web/controller/rest/ArtistsRestControllerTest.java
NilsDammenhayn: How was that comment of you about suggestions that break the code 😉

DanielW1987: Like big L I use suggestions also to make the issue clear. It's your responsibility that the code not breaL :D

webapp/src/main/java/rocks/metaldetector/web/controller/rest/ArtistsRestController.java
DanielW1987: mh, okay...leave it as it is.

PR #172, opened 2020-07-24
fix frontpage layout bug

DanielW1987:
The images no longer run out of the frame and when the browser is minimized, the search bar contracts so that no horizontal scrollbar appears.

The pictures are still not 100% perfect, especially when you switch to the mobile view. But this is a separate issue when we take care of a mobile-optimized layout. For the browser, that's fine for now.


DanielW1987:
[Image scaling on front page](https://trello.com/c/qr7MaOBE/167-image-scaling-on-front-page)

PR #173, opened 2020-07-24
show 400 page on bad request for mvc views and json response for rest endpoints

DanielW1987:


DanielW1987:
[Unpleasant error message when a page that expects URL parameters does not receive any parameters](https://trello.com/c/tQPjdKsx/28-unpleasant-error-message-when-a-page-that-expects-url-parameters-does-not-receive-any-parameters)

PR #174, opened 2020-07-24
sort my artists by artist name

DanielW1987:


DanielW1987:
[[Detector] my artists are not sorted](https://trello.com/c/LfDOk5pq/190-detector-my-artists-are-not-sorted)

PR #176, opened 2020-07-28
Spotify Pagination

hedgehog1833:
One error was that we didn't set the limit in the request. This is now fixed. Now we just have to wait for Spotify to fix their bug 🤷‍♂️


hedgehog1833:
[Search with pagination is broken](https://trello.com/c/tp2OaepG/199-search-with-pagination-is-broken)

PR #177, opened 2020-07-31
Caching of Spotify Authentication Token

hedgehog1833:
This should be everything we have to do to cache the SpotifyAuthenticationToken. I used 59 minutes instead of one hour because the time starts "earlier" on their side. Otherwise it could happen that we think our token is still valid but then Spotify declines it because it is already invalid for them. Counting the request times in we should be safe with one minute backup time I guess.


hedgehog1833:
[[Detector] Caching of Spotify's Auth-Token](https://trello.com/c/bHYAlYtN/194-detector-caching-of-spotifys-auth-token)

hedgehog1833:
Nope, the Integration Test annoyed the hell out of me for the last two days so now I created a trello card for that. It's approved 🤷‍♂️ 😄

PR #180, opened 2020-08-03
Migrate JavaScript from search to TypeScript

DanielW1987:
The search results are now rendered completely by TypeScript.

In addition, we now explicitly point out the Spotify Bug in the result page. This seems to me to be the best strategy.

In the frontend I was able to remove a lot of JavaScript code, including the "old" Pagination plugin.


DanielW1987:
[[Detector] Migrate JavaScript from search to TypeScript](https://trello.com/c/6Y1zjMYR/203-detector-migrate-javascript-from-search-to-typescript)

PR #181, opened 2020-08-06
Homepage rest endpoint

hedgehog1833:
The rest endpoint for the Homepage is implemented with a Service and a Mock. The service is not annotated with @Service as the mock is used for all profiles until the service is complete. It currently only gets the upcoming releases. More is to follow in separate PRs.


webapp/src/main/java/rocks/metaldetector/service/home/HomepageServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        .upcomingReleases(upcomingReleases)
+        .build();
+  }
+
+  private List<ReleaseDto> findUpcomingReleases(List<String> followedArtists, LocalDate now) {
+    if (followedArtists.isEmpty()) {
+      return Collections.emptyList();
+    }
+
+    return releaseService.findReleases(followedArtists, now, now.plusMonths(1))
```
DanielW1987, line 41: I wouldn't want to limit the time too much to ensure that we have enough releases to show. Maybe 6 months.

webapp/src/test/java/rocks/metaldetector/service/home/HomepageServiceImplTest.java
```diff
@@ -0,0 +1,126 @@
...
+        ReleaseDtoFactory.withArtistName("A"), ReleaseDtoFactory.withArtistName("B"),
+        ReleaseDtoFactory.withArtistName("C"), ReleaseDtoFactory.withArtistName("D"),
+        tooMuch);
+    doReturn(List.of(ArtistDtoFactory.createDefault())).when(followArtistService).getFollowedArtistsOfCurrentUser();
+    doReturn(upcomingReleases).when(releaseService).findReleases(any(), any(), any());
+
+    var result = underTest.createHomeResponse();
+
+    // then
+    assertThat(result.getUpcomingReleases().size()).isEqualTo(RESULT_LIMIT);
```
DanielW1987, line 123:
```suggestion
    assertThat(result.getUpcomingReleases().size()).isEqualTo(4);
```

Otherwise the test description will at some point no longer correspond to reality.

hedgehog1833:

PR #184, opened 2020-08-10
Typescript for Homepage

hedgehog1833:
My first TypeScript page 🥳 
I started with the row for recently followed artists.

### DIFF
--- a/webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
+++ b/webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
@@ -1,0 +1,53 @@
+import {LoadingIndicatorService} from "./loading-indicator-service";
+import {AlertService} from "./alert-service";
+import {HomepageResponse} from "../model/homepage-response.model";
+import {AbstractRenderService} from "./abstract-render-service";
+import {Artist} from "../model/artist.model";
+
+export class HomepageRenderService extends AbstractRenderService<HomepageResponse> {
+
+    private readonly artistTemplateElement: HTMLTemplateElement;
+
+    constructor(alertService: AlertService, loadingIndicatorService: LoadingIndicatorService) {
+        super(alertService, loadingIndicatorService);
+        this.artistTemplateElement = document.getElementById("artist-card")! as HTMLTemplateElement;
+    }
+
+    protected getHostElementId(): string {
+        return "home-container";
+    }
+
+    protected onRendering(response: HomepageResponse): void {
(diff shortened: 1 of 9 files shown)


webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    private attachCard(divElement: HTMLDivElement, rowElement: HTMLDivElement) {
+        rowElement.insertAdjacentElement("beforeend", divElement);
+    }
+
+    private createFollowedSinceString(followedSince: Date): string {
+        const now = Date.now();
+        const then = new Date(followedSince).valueOf();
+        const followedForDays = Math.round((now - then) / 1000 / 60 / 60 / 24);
```
DanielW1987, line 50: Isn't there some cool JS function for?

webapp/src/main/resources/templates/frontend/home.html
```diff
@@ -6,261 +6,28 @@
...
-                        <p class="h5 card-title">Alcest</p>
-                    </div>
-                    <div class="card-footer text-muted">
-                        41 days ago
-                    </div>
-                </div>
-            </div>
+        <div id="home-container" class="mb-3">
+            <h2 class="h5 mt-4 mb-2">Recently followed artists</h2>
+            <div id="recently-followed-row" class="row"></div>
```
DanielW1987, line 30: And what does it look like if I don't follow artists at all? Do I only see a headline without content? That shouldn't be.

webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
NilsDammenhayn: Couldn't find one after a short google search...

webapp/src/main/resources/templates/frontend/home.html
NilsDammenhayn: Well I could of course don't display anything if it's empty but wasn't the plan to show a placeholder if it is empty anyway?

DanielW1987: In trello It is specified that we only display the placeholder if there is at least 1 "real" entry. You can fix it with the next PR.

webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
DanielW1987: Maybe you can introduce a date util class with the next PR.

hedgehog1833:

PR #186, opened 2020-08-15
Prepare backend to fetch releases with page and size parameter

DanielW1987:


DanielW1987:
[Revision of releases page](https://trello.com/c/2aKFnPXh/91-revision-of-releases-page)

PR #191, opened 2020-08-23
Recent releases in HomepageService

hedgehog1833:
Recent releases for the last six months are provided by HomepageService


webapp/src/main/java/rocks/metaldetector/service/home/HomepageServiceImpl.java
```diff
@@ -27,20 +27,33 @@ public HomepageResponse createHomeResponse() {
     List<String> followedArtists = followArtistService.getFollowedArtistsOfCurrentUser()
         .stream().map(ArtistDto::getArtistName).collect(Collectors.toList());
 
-    List<ReleaseDto> upcomingReleases = findUpcomingReleases(followedArtists);
+    LocalDate now = LocalDate.now();
+    PageRequest pageRequest = new PageRequest(1, RESULT_LIMIT);
+
+    List<ReleaseDto> upcomingReleases = findUpcomingReleases(followedArtists, now, pageRequest);
+    List<ReleaseDto> recentReleases = findRecentReleases(followedArtists, now, pageRequest);
+
```
DanielW1987, line 35: I think the concept here is not very sustainable ;) Especially if the other home page data queries are added here as well. This will be difficult or very extensive / complex even during the tests in the setup.

My idea would be the following:
- `../service/summary/SummaryService`

The `SummaryService` uses the following data collectors:
- `../service/summary/UpcomingReleasesCollector`
- `../service/summary/RecentReleasesCollector`
- `../service/summary/WhateverCollector`

I like `Collector`, because they use other services to get the data.
`SummaryService` is tbd - I think it's clearer than `HomepageService`.

DanielW1987 (approved):
Since all this is still very lively, you are welcome to address my change request in the follow-up PR.

webapp/src/main/java/rocks/metaldetector/service/home/HomepageServiceImpl.java
hedgehog1833: Sounds good! Testing already annoyed me now just as the name for the HomepageService :-D

hedgehog1833:

PR #192, opened 2020-08-24
Recent releases on homepage

hedgehog1833:
The recent releases are loaded dynamically on the homepage. Also a util class for the time range calculation is added.


DanielW1987 (approved):
Could you maybe add a checklist in the Trello cards so I know what other things you will address?

For example, I'm still missing the placeholder card and that a row will not be rendered at all if no elements are present.

hedgehog1833:

PR #193, opened 2020-08-25
SummaryService returns top followed artists

hedgehog1833:
- Refactoring of HomepageService (now `SummaryService`) with collector classes
- Custom query to get top followed artists
- SummaryService returns top followed artists


hedgehog1833:

persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/ArtistRepository.java
```diff
@@ -16,4 +18,11 @@
 
   boolean existsByExternalIdAndSource(String externalId, ArtistSource source);
 
+  @Query(value = "select a.id, a.created_by, a.created_date, a.last_modified_by, a.last_modified_date, a.artist_name, a.external_id, a.source, a.thumb " +
+                 "from artists a left join users_followed_artists ufa on a.id = ufa.artist_id " +
+                 "group by a.id, a.created_by, a.created_date, a.last_modified_by, a.last_modified_date, a.artist_name, a.external_id, a.source, a.thumb " +
+                 "order by count(a.id) desc " +
+                 "limit :limit",
+         nativeQuery = true)
+  List<ArtistEntity> findTopArtists(@Param("limit") int limit);
```
DanielW1987, line 27: Don't think we need the whole ArtistEntity back. Otherwise we have a very hidden maintenance effort here, which we always have to adjust when we change the ArtistEntity.

Think we need something like this: <https://www.baeldung.com/jpa-queries-custom-result-with-aggregation-functions>

Good enough for now, but we should address this in future.

DanielW1987 (approved):

PR #195, opened 2020-08-25
Dependency updates

hedgehog1833:
Updates for spring boot and several other dependencies. I downgraded rest-assured from 4.3.0 to 4.2.0 to get rid of the groovy dependency. Apparently this is an issue in 4.3.0 (and 4.3.1) that also exists in other Java Version than 14. So let's just wait here.
I didn't manage to get Mockito updated so it's still 3.1.0. I'll create a dependency card for these two points.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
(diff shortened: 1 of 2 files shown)


hedgehog1833:
[Check Detector dependencies](https://trello.com/c/bJI3GDuL/162-check-detector-dependencies)

DanielW1987 (approved):

PR #197, opened 2020-08-27
SummaryServiceImpl collects the community's most expected releases

hedgehog1833:


hedgehog1833:

webapp/src/main/java/rocks/metaldetector/service/summary/ReleaseCollector.java
```diff
@@ -17,36 +17,43 @@
...
 
-    return collectReleases(timeRange);
+    return collectReleases(followedArtists, timeRange);
+  }
+
+  public List<ReleaseDto> collectUpcomingReleases(List<ArtistDto> artists) {
+    LocalDate now = LocalDate.now();
+    TimeRange timeRange = new TimeRange(now, now.plusMonths(TIME_RANGE_MONTHS));
+
+    return collectReleases(artists, timeRange);
```
DanielW1987, line 39: The most followed artists are determined outside this class and the followed artists of the user are determined within this class. This makes it a bit more difficult to understand the sense of these methods. 

I would let the SummaryService determine both (most followed artists and artists of the user). Then you only need one method here.

webapp/src/main/java/rocks/metaldetector/service/summary/ReleaseCollector.java
```diff
@@ -17,36 +17,43 @@
...
   }
 
   public List<ReleaseDto> collectRecentReleases() {
+    List<ArtistDto> followedArtists = followArtistService.getFollowedArtistsOfCurrentUser();
```
DanielW1987, line 43: This could also be a parameter...

DanielW1987 (approved):

PR #198, opened 2020-08-27
Typescript for favorite community artists, placeholder and some more stuff

hedgehog1833:
This PR contains a few changes:
- Favorite community artists are fetched and displayed on homepage
- Placeholder card is shown if there are only three cards in a release row
- Only one release row is created if not enough results are present for two rows
- If this row would have more that four elements the last ones are cut

There are still some open points here. One would be that there is no placeholder if there are not enough recently followed artists. This is because we cannot use the artist placeholder because the card would have a different height as the artist cards have one row less text.
Also play around a little with the mock and put more or less releases in the list to see what happens. The placeholder layout of course is open for discussion.


hedgehog1833:

DanielW1987 (commented):
Can you please resize the question mark to reduce traffic? The current homepage pictures are 300x300 in size (more or less).

webapp/src/main/resources/static/ts/src/service/date-service.ts
```diff
@@ -14,4 +16,17 @@ export class DateService {
...
+        const date2AsNumber = release2.releaseDate.valueOf();
+
+        if (date1AsNumber < date2AsNumber) {
+            return -1;
+        }
+        if (date2AsNumber < date1AsNumber) {
+            return 1;
+        }
+        return 0;
+    }
```
DanielW1987, line 31: I would rather sort in the backend. Is there something to be said against that?

DanielW1987: If it's not possible why not just simple return

```typescript
return release1.releaseDate.valueOf() - release2.releaseDate.valueOf();
```

webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
```diff
@@ -24,9 +25,31 @@ export class HomepageRenderService extends AbstractRenderService<HomepageRespons
     }
 
     protected onRendering(response: HomepageResponse): void {
-        this.renderUpcomingReleasesRow(response);
-        this.renderRecentReleasesRow(response);
-        this.renderRecentlyFollowedRow(response);
+        if (response.upcomingReleases.length >= this.CARDS_PER_ROW - 1 &&
+            response.recentReleases.length >= this.CARDS_PER_ROW - 1) {
```
DanielW1987, line 29: why do we need `>=` and `-1` in combination here...?

DanielW1987: Maybe it's clearer to use simply 3 or introduce a new constant. 
Maybe in the future `CARDS_PER_ROW` is 8.

webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
```diff
@@ -123,4 +198,11 @@ export class HomepageRenderService extends AbstractRenderService<HomepageRespons
         this.hostElement.insertAdjacentElement("beforeend", recentlyFollowedRowElement);
         return recentlyFollowedRowElement;
     }
+
+    private insertPlaceholder(elementCount: number, rowElement: HTMLDivElement): void {
+        if (elementCount == this.CARDS_PER_ROW - 1) {
+            const placeholderDivElement = this.renderPlaceholderCard();
+            this.attachCard(placeholderDivElement, rowElement);
+        }
+    }
```
DanielW1987, line 207: Do we want a placeholder if it is 3 or less, isn't it?

webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
```diff
@@ -39,40 +62,76 @@ export class HomepageRenderService extends AbstractRenderService<HomepageRespons
             releaseDateElement.innerHTML = this.createReleasedBeforeString(release.releaseDate);
             this.attachCard(releaseDivElement, recentReleasesRowElement);
         });
+
+        if (response.recentReleases.length == this.CARDS_PER_ROW - 1) {
+            const placeholderDivElement = this.renderPlaceholderCard();
+            this.attachCard(placeholderDivElement, recentReleasesRowElement);
+        }
+        this.insertPlaceholder(response.recentReleases.length, recentReleasesRowElement);
```
DanielW1987, line 70: this looks like a mistake

webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
```diff
@@ -39,40 +62,76 @@ export class HomepageRenderService extends AbstractRenderService<HomepageRespons
...
+                    const releaseDivElement = this.renderReleaseCard(release);
+                    const releaseDateElement = releaseDivElement.querySelector("#release-date") as HTMLDivElement;
+                    releaseDateElement.innerHTML = this.createReleasedInString(release.releaseDate);
+                    this.attachCard(releaseDivElement, releasesRow);
+                }
+            });
+
+            const elementCount = response.recentReleases.length + response.upcomingReleases.length;
+            this.insertPlaceholder(elementCount, releasesRow);
+        }
```
DanielW1987, line 98: It should be easier if you simply add the releases from the two lists to a new list first and then just iterate over that list. First add the upcoming releases and after that the recent releases. You can break the for loop after 4.

DanielW1987 (changes requested):

webapp/src/main/resources/static/ts/src/service/date-service.ts
hedgehog1833: The problem is that the sorting only has to be done when there are so few results that we only show one row. And in one row it looks strange if both - the recent and the upcoming - releases are sorted in ascending order, as we would have something like "release A 2 days ago, release B 5 days ago, release C in 1 day, release D in 10 days". In this case I would prefer having the chronological order for all releases.
And I think that the backend shouldn't care about something like that so yeah, ordering here seems right. But also your second suggestions looks better so I'll apply that 👍

webapp/src/main/resources/static/ts/src/service/homepage-render-service.ts
hedgehog1833: Yes. I still struggle with the idea of having always full rows of 4 cards. The idea to prevent that (using one row if there are too few releases) does not always work out perfectly.. But yes, I set the value to `elementCount < this.CARDS_PER_ROW`

webapp/src/main/resources/static/ts/src/service/date-service.ts
hedgehog1833: As just discussed: ordering is now in the backend 😂

DanielW1987 (approved):

PR #201, opened 2020-08-29
Deleted artists from release requests, changed postmapping to getmapping

ReginaSchuster:
I deleted the artists lists from the findAllReleases and findReleases methods and the corresponding classes. Also, I changed the @PostMapping to a @GetMapping for both methods with the parameters from the ReleasesRequest and the PaginatedReleasesRequest (classes still have to be deleted). Draft does not compile due to tests.


DanielW1987 (commented):
It is OK to remove the `artists' from the requests of the `ReleasesRestController'. But we still need the artists in the `ReleaseService`, otherwise we can't determine the releases of the user's favorite artists ;)

From the controller you should pass on an empty list for now. An empty list means that the `ReleaseService` will ask for the releases of all artists.

DanielW1987 (approved):

PR #202, opened 2020-08-29
show limited amount of pages

DanielW1987:
The pagination component now shows dot items on more than 10 pages.


DanielW1987:
[Pagination: Show limited amount of pages](https://trello.com/c/Sh6LG713/198-pagination-show-limited-amount-of-pages)

hedgehog1833 (approved):
GEFN but there is still an issue when there are many pages. Spotify's api documentation states that the offset when searching has a maximum value of 2000 ([see here](https://developer.spotify.com/documentation/web-api/reference/search/search/)). For offsets > 2000 we get 404.

DanielW1987:
> GEFN but there is still an issue when there are many pages. Spotify's api documentation states that the offset when searching has a maximum value of 2000 ([see here](https://developer.spotify.com/documentation/web-api/reference/search/search/)). For offsets > 2000 we get 404.

For what artist I have to search to reproduce this?

hedgehog1833:
Just type in incomplete name. I started with `Ope` or just type `a`

DanielW1987:
will have a look at it...

PR #203, opened 2020-08-31
Spring data projection

hedgehog1833:
An interface for top artists is introduced and with the help of spring data projection populated from the custom sql query. So now we don't create full ArtistEntities on getting the top artists but just have name and thumb.


hedgehog1833:
[[Detector] Replace custom SQL](https://trello.com/c/mLJbLacr/213-detector-replace-custom-sql)

DanielW1987 (approved):

PR #204, opened 2020-09-02
dateTo cannot be set without dateFrom in ReleasesRequests

hedgehog1833:
I put another method in the `WithTimeRangeValidation` interface, please check the naming. I wasn't so sure there...


hedgehog1833:
[Bad Request if we pass only dateFrom](https://trello.com/c/yg3ecF7g/218-bad-request-if-we-pass-only-datefrom)

DanielW1987 (approved):

PR #206, opened 2020-09-10
Refactoring of ArtistsService

ReginaSchuster:
Renamed ArtistsService to ArtistService, factored out all search-related methods into new classes ArtistSearchService and ArtistSearchServiceImpl


webapp/src/test/java/rocks/metaldetector/web/controller/rest/ArtistsRestControllerTest.java
```diff
@@ -73,7 +77,7 @@ void setUp() {
 
     @AfterEach
     void tearDown() {
-      reset(artistsService);
+      reset(artistSearchService);
```
DanielWagner87, line 80: here we should reset all mocks of the test class.

webapp/src/test/java/rocks/metaldetector/service/artist/ArtistServiceImplTest.java
```diff
@@ -45,7 +45,7 @@
 import static rocks.metaldetector.testutil.DtoFactory.SpotifyArtistSearchResultDtoFactory;
 
 @ExtendWith(MockitoExtension.class)
-class ArtistsServiceImplTest implements WithAssertions {
+class ArtistServiceImplTest implements WithAssertions {
```
DanielWagner87, line 48: We should also split up the tests into `ArtistServiceImplTest` and `ArtistSearchServiceImplTest`. Convention is that we only test one class per test class and the instance of the class under test is called `underTest`

DanielWagner87 (commented):
So far, so good. Two small things...

DanielW1987 (approved):
lgtm

PR #209, opened 2020-09-12
Revision of releases page

DanielW1987:
Finally...

We can now see all upcoming releases. The Trello card contains a number of to-dos. For example, the right filter bar has no function yet.


DanielW1987:
[Revision of releases page](https://trello.com/c/2aKFnPXh/91-revision-of-releases-page)

support/src/main/java/rocks/metaldetector/support/EnumPrettyPrinter.java
```diff
@@ -3,10 +3,24 @@
 import org.apache.commons.text.WordUtils;
 import org.springframework.stereotype.Service;
 
+import java.util.List;
+
 @Service
 public class EnumPrettyPrinter {
 
+  private static final List<String> EXCEPTIONS = List.of(
```
hedgehog1833, line 11: The variable name really got me confused... but I don't have better ideas 🤷

hedgehog1833:
Ah and I got a lot of 400 errors for the images but I guess that's my local database, haven't checked yet

DanielW1987:
> Ah and I got a lot of 400 errors for the images but I guess that's my local database, haven't checked yet

I have the same error, it's your local database. Later we need something to use also images from local disc.

support/src/main/java/rocks/metaldetector/support/EnumPrettyPrinter.java
DanielW1987: `EXCEMPTION`?

support/src/test/java/rocks/metaldetector/support/EnumPrettyPrinterTest.java
```diff
@@ -34,6 +34,17 @@ void should_transform_fully_capitalized(String givenValue, String expectedValue)
     assertThat(result).isEqualTo(expectedValue);
   }
 
+  @ParameterizedTest(name = "Should not transform <{0}>")
```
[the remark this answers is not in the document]
DanielW1987: ...to be prepared to future enhancements of the exemption list.

support/src/main/java/rocks/metaldetector/support/EnumPrettyPrinter.java
hedgehog1833: Yeah I guess that's better

hedgehog1833 (approved):
Niiice 🤘

PR #210, opened 2020-09-17
Changed followed-by statement to account for 0 and 1 follower

ReginaSchuster:


webapp/src/main/resources/static/ts/src/service/search-render-service.ts
```diff
@@ -118,11 +118,31 @@ export class SearchRenderService extends AbstractRenderService<SearchResponse> {
...
-        followInfoElement.textContent = `Followed by ${entry.metalDetectorFollower} users`;
+        followInfoElement.textContent = this.determineFollowerAmountStatement(entry.metalDetectorFollower);
 
         return topSearchResultDivElement;
     }
 
+    private determineFollowerAmountStatement(metalDetectorFollower: number): string {
+
+        let followerAmountStatement: string;
+        followerAmountStatement = "";
```
DanielW1987, line 129: can be shorten to 

```
let followerAmountStatement = "";
```

DanielW1987 (approved):

PR #213, opened 2020-10-03
Persist followedSince Date and show in frontend

DanielW1987:
There is now a 'FollowActionEntity' which actually feels quite good. All Follow Actions are now persisted in a separate table.

### DIFF
--- a/persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/ArtistRepository.java
+++ b/persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/ArtistRepository.java
@@ -19,7 +19,7 @@ public interface ArtistRepository extends JpaRepository<ArtistEntity, Long> {
   boolean existsByExternalIdAndSource(String externalId, ArtistSource source);
 
   @Query(value = "select a.artist_name as artistName, a.thumb as thumb " +
-                 "from artists as a left join users_followed_artists as ufa on a.id = ufa.artist_id " +
+                 "from artists as a left join follow_actions as fa on a.id = fa.artist_id " +
                  "group by a.artist_name, a.thumb " +
                  "order by count(a.id) desc " +
                  "limit :limit",
@@ -27,7 +27,7 @@ public interface ArtistRepository extends JpaRepository<ArtistEntity, Long> {
   List<TopArtist> findTopArtists(@Param("limit") int limit);
 
   @Query(value = "select count(a.external_id) " +
-                 "from artists a left join users_followed_artists ufa on a.id = ufa.artist_id " +
+                 "from artists a left join follow_actions fa on a.id = fa.artist_id " +
                  "where a.external_id = :externalId",
          nativeQuery = true)
   int countArtistFollower(@Param("externalId") String externalId);
--- a/persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/FollowActionEntity.java
+++ b/persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/FollowActionEntity.java
@@ -1,6 +1,7 @@
 package rocks.metaldetector.persistence.domain.artist;
 
 import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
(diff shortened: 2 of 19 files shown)


DanielW1987:
[Persist followedSince Date](https://trello.com/c/zsTUDTak/217-persist-followedsince-date)

persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/FollowActionEntity.java
```diff
@@ -0,0 +1,37 @@
...
+  @NonNull
+  private UserEntity user;
+
+  @OneToOne(targetEntity = ArtistEntity.class)
+  @JoinColumn(nullable = false, name = "artist_id")
+  @NonNull
+  private ArtistEntity artist;
+
+  @Builder
+  public FollowActionEntity(@NonNull UserEntity user, @NonNull ArtistEntity artist) {
```
NilsDammenhayn, line 33: Why not the `@AllArgsConstructor`? Because of NonNull?

persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/ArtistRepository.java
```diff
@@ -19,15 +19,15 @@
...
                  "limit :limit",
          nativeQuery = true)
   List<TopArtist> findTopArtists(@Param("limit") int limit);
 
   @Query(value = "select count(a.external_id) " +
-                 "from artists a left join users_followed_artists ufa on a.id = ufa.artist_id " +
+                 "from artists a left join follow_actions fa on a.id = fa.artist_id " +
                  "where a.external_id = :externalId",
          nativeQuery = true)
   int countArtistFollower(@Param("externalId") String externalId);
```
NilsDammenhayn, line 33: I think we can replace this query now with something like `countFollowActionEntitiesByArtist(ArtistEntity artistEntity)` in the `FollowActionEntityRepository`.

persistence/src/test/java/rocks/metaldetector/persistence/domain/artist/FollowActionRepositoryTest.java
```diff
@@ -0,0 +1,76 @@
...
+import rocks.metaldetector.persistence.domain.user.UserEntity;
+import rocks.metaldetector.persistence.domain.user.UserFactory;
+import rocks.metaldetector.persistence.domain.user.UserRepository;
+
+import java.util.List;
+
+import static rocks.metaldetector.persistence.domain.artist.ArtistSource.DISCOGS;
+import static rocks.metaldetector.persistence.domain.artist.ArtistSource.SPOTIFY;
+
+class FollowActionRepositoryTest extends BaseDataJpaTest implements WithAssertions, WithIntegrationTestConfig {
```
NilsDammenhayn, line 20: Maybe one small test for `exists` if you are already testing the other methods? ;)

webapp/src/test/java/rocks/metaldetector/service/artist/FollowArtistServiceImplTest.java
```diff
@@ -286,79 +290,177 @@ void unfollow_should_remove_artist_from_user() {
...
+    // when
+    underTest.isCurrentUserFollowing(EXTERNAL_ID, ARTIST_SOURCE);
+
+    // then
+    verify(currentPublicUserIdSupplier, times(1)).get();
+    verify(userRepository, times(1)).findByPublicId(PUBLIC_USER_ID);
+  }
+
+  @Test
+  @DisplayName("isCurrentUserFollowing(): should fetch user entity")
```
NilsDammenhayn, line 312: Copy+Paste error. Should be something like "should throw exception"

NilsDammenhayn (commented):
Very nice, just some small things ❤️

persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/FollowActionEntity.java
DanielW1987: i'm using `@AllArgsConstructor` now

persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/ArtistRepository.java
DanielW1987: I would actually like to have that in the `ArtistRepository`.

Furthermore, we only have the External ID available at the calling point. We would have to map this to an ArtistEntity again.

NilsDammenhayn (approved):

PR #214, opened 2020-10-05
Blog backend initialsetup

ReginaSchuster:
Created blogpost entity, dummy posts and Rest endpoint to get all posts


ReginaSchuster:
This was compiling fine locally, but showed a Maven error in the PR. After merging the master branch into this branch and re-running it locally, I get the following error: "java: error: release version 15 not supported". Couldn't figure it out right away and will look into this again.

persistence/src/main/java/rocks/metaldetector/persistence/domain/blogpost/BlogpostEntity.java
```diff
@@ -0,0 +1,38 @@
...
+import lombok.*;
+
+import rocks.metaldetector.persistence.domain.BaseEntity;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.validation.constraints.Size;
+
+@Getter
+@NoArgsConstructor
```
hedgehog1833, line 12: This constructor does not have to be accessible from anywhere (an import is needed for this suggestion to work).
```suggestion
@NoArgsConstructor(access = AccessLevel.PROTECTED) // for hibernate and model mapper
```

webapp/src/main/java/rocks/metaldetector/service/blogpost/BlogpostServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+import rocks.metaldetector.persistence.domain.blogpost.BlogpostEntity;
+import rocks.metaldetector.persistence.domain.blogpost.BlogpostRepository;
+
+import java.util.List;
+
+@Service
+@Slf4j
```
hedgehog1833, line 12: I wouldn't add this if you're not logging anything.

persistence/src/main/java/rocks/metaldetector/persistence/domain/blogpost/BlogpostEntity.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Column(name = "slug", nullable = false, unique = true)
+    @Size(min = 2, max = 255)
+    private String slug;
+
+    @Column(columnDefinition="TEXT", name = "text", nullable = false)
+    private String text;
+
+    @Column(name = "draft_flag", nullable = false)
+    private boolean draftFlag;
```
hedgehog1833, line 28: I think just `draft` would be better for a name. Otherwise the getter will be `isDrafFlag()` which sounds a bit weird.

persistence/pom.xml
```diff
@@ -34,6 +34,14 @@
             <artifactId>metal-detector-support</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <!-- Validation -->
```
hedgehog1833, line 38: This dependency is not needed, see my comment in `BlogpostEntity.java`

persistence/src/main/java/rocks/metaldetector/persistence/domain/blogpost/BlogpostEntity.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Builder
+    public BlogpostEntity(@NonNull String title, @NonNull String slug, @NonNull String text,
```
hedgehog1833, line 31: The `@NotNull` is not necessary here. It is ensured on database level that these values are not null. Also for consistency reasons I would say that we omit them as we never used it so far.

persistence/src/main/java/rocks/metaldetector/persistence/domain/blogpost/BlogpostEntity.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.persistence.Entity;
+import javax.validation.constraints.Size;
+
+@Getter
+@NoArgsConstructor
+@Entity(name = "blogposts")
+public class BlogpostEntity extends BaseEntity {
+
+    @Column(name = "title", nullable = false, unique = true)
+    @Size(min = 2, max = 255)
```
hedgehog1833, line 17: I guess using the `@Size` annotation would work, but the additional Dependency is not necessary here. `@Column` has a `length` parameter that can be overwritten if needed (default is 255) and the minimum value can be left out in my opinion.
Please also check the other member.

persistence/src/main/java/rocks/metaldetector/persistence/domain/blogpost/BlogpostEntity.java
```diff
@@ -0,0 +1,38 @@
...
+    private String title;
+
+    @Column(name = "slug", nullable = false, unique = true)
+    @Size(min = 2, max = 255)
+    private String slug;
```
hedgehog1833, line 22: What's a slug in this context? I thought a slug is this: 🐌 😅  I guess it is the English term for the short version of an article?

persistence/src/main/java/rocks/metaldetector/persistence/config/bootstrap/DefaultDatabaseInitializer.java
```diff
@@ -107,4 +109,46 @@ private void createAndFollowArtists() {
...
+            .draftFlag(true)
```
hedgehog1833, line 146: Maybe add one example that is no draft?

webapp/src/test/java/rocks/metaldetector/service/blogpost/BlogpostServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+import rocks.metaldetector.persistence.domain.blogpost.BlogpostEntity;
+import rocks.metaldetector.persistence.domain.blogpost.BlogpostRepository;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
+public class BlogpostServiceTest implements WithAssertions {
```
hedgehog1833, line 21: Our naming schema so far is naming the tests for service implementations are the certain class, so this would be `BlogpostServiceImplTest`. That's nicer when you have multiple implementations (or mocks).

hedgehog1833 (changes requested):
Looks fine, just some smaller remarks 👍  And a test for the `BlogpostRestController` is missing 😉

PR #216, opened 2020-10-14
User authorization Part II

hedgehog1833:
After being redirected to the profile page the state is checked and the code is exchanged for and authorization token, which is then persisted.
What's still missing: clicking again just fetches a new token with a new code instead of checking if an non-expired token exists and using the refresh token if needed. This comes later PRs.


hedgehog1833:
[Spotify Artist Import](https://trello.com/c/jTFeZAZc/220-spotify-artist-import)

persistence/src/main/java/rocks/metaldetector/persistence/domain/spotify/SpotifyAuthorizationEntity.java
```diff
@@ -20,12 +20,21 @@
   @Column(name = "state", nullable = false, updatable = false)
   private String state;
 
-  @Column(name = "access_token", unique = true)
+  @Column(name = "access_token")
   private String accessToken;
 
-  @Column(name = "refresh_token", unique = true)
+  @Column(name = "refresh_token")
```
DanielW1987, line 26: Why do they not have to be unique anymore?

webapp/src/main/java/rocks/metaldetector/service/spotify/SpotifyUserAuthorizationServiceImpl.java
```diff
@@ -34,4 +38,36 @@ public String prepareAuthorization() {
 
     return spotifyService.getSpotifyAuthorizationUrl() + state;
   }
+
+  @Override
+  @Transactional
+  @Async
```
DanielW1987, line 44: Should this really be asynchronous at this point? The user should not be allowed to continue until the token is persisted. Otherwise he could theoretically want to import his artists without having a valid token?

webapp/src/main/java/rocks/metaldetector/web/controller/mvc/CustomErrorController.java
```diff
@@ -19,7 +19,7 @@
 
   @Override
   public String getErrorPath() {
-    return Endpoints.ERROR;
+    return null;
```
DanielW1987, line 22: Why? And how does Spring recognize this controller as an error controller?

webapp/src/main/java/rocks/metaldetector/service/spotify/SpotifyUserAuthorizationServiceImpl.java
hedgehog1833: My idea was that the user does not have to wait for the calls to be made and the token to be persisted until the page is loaded. And yes in theory it could be possible that the token is not persisted when importing is started....

DanielW1987: Ok, and how do you inform the user about errors?

persistence/src/main/java/rocks/metaldetector/persistence/domain/spotify/SpotifyAuthorizationEntity.java
hedgehog1833: It seemed unnecessary to me 🤷‍♂️ One user has one authentication entity and the tokens won't be the same for different users anyway. So why have this here..

webapp/src/main/java/rocks/metaldetector/web/controller/mvc/CustomErrorController.java
hedgehog1833: It is not needed anymore. The path is now defined in the `application.yml` (since some spring version that I forgot). The method is already deprecated in the interface but as it still exists the method cannot be removed. And this way we see that it is not needed anymore.

DanielW1987: Can we remove the interface?

webapp/src/main/java/rocks/metaldetector/service/spotify/SpotifyUserAuthorizationServiceImpl.java
hedgehog1833: But yeah maybe a user would not really perceive the time as too long so it would not be an issue to make this synchronous..

persistence/src/main/java/rocks/metaldetector/persistence/domain/spotify/SpotifyAuthorizationEntity.java
DanielW1987: To avoid programmatic mistakes done by humans?

DanielW1987: It's ok for me...

webapp/src/main/java/rocks/metaldetector/web/controller/mvc/CustomErrorController.java
DanielW1987: Can you check the removal of that interface...?

hedgehog1833: Doesn't work 😕  The application won't start then because of "ambiguous mappings`... Guess it has to stay

webapp/src/main/java/rocks/metaldetector/service/spotify/SpotifyUserAuthorizationServiceImpl.java
hedgehog1833: > Ok, and how do you inform the user about errors?

Currently not at all 😅  I'll think about something...

DanielW1987 (approved):

PR #217, opened 2020-10-15
Recently followed artists on homepage and mock deactivation

hedgehog1833:
The recently followed artists are now shown on the homepage. The SummaryService's mock is deactivated and the number of followers added to the community's favorite artists.
It might be good to activate the mock also for the "default" profile, as we probably won't have that much content locally so the homepage looks quite barren. What do you think?

### DIFF
--- a/webapp/src/test/java/rocks/metaldetector/service/summary/ArtistCollectorTest.java
+++ b/webapp/src/test/java/rocks/metaldetector/service/summary/ArtistCollectorTest.java
@@ -8,15 +8,30 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import rocks.metaldetector.persistence.domain.artist.ArtistEntity;
 import rocks.metaldetector.persistence.domain.artist.ArtistRepository;
+import rocks.metaldetector.persistence.domain.artist.FollowActionEntity;
+import rocks.metaldetector.persistence.domain.artist.FollowActionRepository;
 import rocks.metaldetector.persistence.domain.artist.TopArtist;
+import rocks.metaldetector.persistence.domain.user.UserEntity;
+import rocks.metaldetector.persistence.domain.user.UserRepository;
+import rocks.metaldetector.security.CurrentPublicUserIdSupplier;
+import rocks.metaldetector.service.artist.ArtistEntityFactory;
 import rocks.metaldetector.service.artist.ArtistTransformer;
+import rocks.metaldetector.service.user.UserEntityFactory;
+import rocks.metaldetector.support.exceptions.ResourceNotFoundException;
 import rocks.metaldetector.testutil.DtoFactory.ArtistDtoFactory;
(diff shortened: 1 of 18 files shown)


hedgehog1833:
[Endpoint for start page infos](https://trello.com/c/hgvQjrHB/208-endpoint-for-start-page-infos)

DanielW1987:
> It might be good to activate the mock also for the "default" profile, as we probably won't have that much content locally so the homepage looks quite barren. What do you think?

But then we can only test it properly on Preview. With the default profile I would like to have the real behaviour.

webapp/src/test/java/rocks/metaldetector/service/summary/SummaryServiceImplTest.java
```diff
@@ -160,4 +170,18 @@ void test_most_expected_releases_returned() {
     // then
     assertThat(result.getMostExpectedReleases()).isEqualTo(releases);
   }
+
+  @Test
+  @DisplayName("")
```
DanielW1987, line 175: nice display name...

webapp/src/test/java/rocks/metaldetector/service/summary/ArtistCollectorTest.java
```diff
@@ -81,4 +110,145 @@ void test_artist_dtos_are_returned() {
...
+  void test_result_limited() {
+    // given
+    var artist = ArtistEntityFactory.withExternalId("1");
+    var userFollowsArtist1 = FollowActionEntity.builder().user(userEntity).artist(artist).build();
+    var userFollowsArtist2 = FollowActionEntity.builder().user(userEntity).artist(artist).build();
+    var userFollowsArtist3 = FollowActionEntity.builder().user(userEntity).artist(artist).build();
+    var userFollowsArtist4 = FollowActionEntity.builder().user(userEntity).artist(artist).build();
+    var userFollowsArtist5 = FollowActionEntity.builder().user(userEntity).artist(artist).build();
+    var followActionEntities = List.of(userFollowsArtist1, userFollowsArtist2, userFollowsArtist3, userFollowsArtist4, userFollowsArtist5);
+    followActionEntities.forEach(action -> action.setCreatedDateTime(Date.from(Instant.now())));
```
DanielW1987, line 226:
```suggestion
    var followActionEntities = IntStream.rangeClosed(1, RESULT_LIMIT + 1)
            .mapToObj(value -> FollowActionEntity.builder().user(userEntity).artist(artist).build())
            .peek(action -> action.setCreatedDateTime(Date.from(Instant.now())))
            .collect(Collectors.toList());
```

hedgehog1833:
> But then we can only test it properly on Preview. With the default profile I would like to have the real behaviour.
Yeah sounds sensible, so maybe you should add some default data at some point.

DanielW1987:
> > But then we can only test it properly on Preview. With the default profile I would like to have the real behaviour.
> 
> Yeah sounds sensible, so maybe you should add some default data at some point.

Yeah sounds sensible, so maybe you should add some default data at some point ;)

DanielW1987 (approved):

PR #219, opened 2020-10-25
Spotify artist import

hedgehog1833:
It is now possible to import artists from Spotify. The import takes all albums the user has liked on Spotify and then fetches the artists. The import button is currently placed on the MyArtists page, this is open for debate. One thing I'd like to have in the future is a mechanism that shows a list of artists to import in the frontend to check/uncheck and takes only those that you want. I'm also thinking of other ways to import, not by album but by really followed artists. But as I use Spotify more the way that I save albums rather than follow artists, that was may first approach here.
Regarding the "workflow": it is currently possible to click on import before authorizing, which obviously does not work. I am not really sure what the best way to prevent that would be. Maybe we don't need an authorization button at all and this is done automatically when clicking on import 🤷‍♂️ But this PR is big enough already, so one step after the other :)


hedgehog1833:
[Spotify Artist Import](https://trello.com/c/jTFeZAZc/220-spotify-artist-import)

DanielW1987:
@hedgehog1833 
You have conflits here.

DanielW1987:
Authorization and import should be done together and not separately. I would have liked to see the synchronization with Spotify in the profile. The reason for this is that the synchronization becomes more complex.

But these are all topics for later PRs.

webapp/src/main/java/rocks/metaldetector/service/artist/ArtistServiceImpl.java
```diff
@@ -37,4 +39,21 @@ public boolean existsArtistByExternalId(String externalId, ArtistSource source)
...
+  }
+
+  @Override
+  public List<String> findNewArtistIds(List<String> artistIds) {
+    List<String> existingArtistIds = artistRepository.findAllByExternalIdIn(artistIds).stream()
+        .map(ArtistEntity::getExternalId)
+        .collect(Collectors.toList());
+    return artistIds.stream()
+        .filter(id -> !existingArtistIds.contains(id))
+        .collect(Collectors.toList());
```
DanielW1987, line 57:
```suggestion
    artistRepository.findAllByExternalIdIn(artistIds).stream()
        .map(ArtistEntity::getExternalId)
        .forEach(artistIds::remove);

    return artistIds;
```

spotify/src/main/java/rocks/metaldetector/spotify/facade/SpotifyServiceImpl.java
```diff
@@ -27,45 +35,74 @@
...
   @Override
   public SpotifyUserAuthorizationDto getAccessToken(String code) {
-    SpotifyUserAuthorizationResponse response = spotifyAuthorizationClient.getUserAuthorizationToken(code);
-    return spotifyUserAuthorizationTransformer.transform(response);
+    SpotifyUserAuthorizationResponse response = authorizationClient.getUserAuthorizationToken(code);
+    return userAuthorizationTransformer.transform(response);
+  }
+
+  @Override
+  public List<SpotifyAlbumDto> importAlbums(String token) {
```
DanielW1987, line 86: That's not really an import at this point. It's more a fetch of liked albums or somethin like that.

spotify/src/main/java/rocks/metaldetector/spotify/client/SpotifyImportClient.java
```diff
@@ -0,0 +1,8 @@
+package rocks.metaldetector.spotify.client;
+
+import rocks.metaldetector.spotify.api.imports.SpotfiyAlbumImportResult;
+
+public interface SpotifyImportClient {
+
+  SpotfiyAlbumImportResult importAlbums(String token, int offset);
+}
```
DanielW1987, line 8: The import term is misleading here. The import is made at a later step.

webapp/src/main/java/rocks/metaldetector/service/spotify/SpotifyArtistImportServiceImpl.java
```diff
@@ -0,0 +1,68 @@
...
+        .collect(Collectors.toList());
+  }
+
+  private void persistNewArtists(List<String> newArtistsIds) {
+    List<SpotifyArtistDto> spotifyArtistDtos = new ArrayList<>();
+    int totalPages = (int) Math.ceil((double) newArtistsIds.size() / (double) PAGE_SIZE);
+    for (int i = 1; i <= totalPages; i++) {
+      List<String> idsPerPage = slicingService.slice(newArtistsIds, i, PAGE_SIZE);
+      spotifyArtistDtos.addAll(spotifyService.searchArtistsByIds(idsPerPage));
+    }
```
DanielW1987, line 65: The for loop should be done within `SpotifyServiceImpl`, because there is already a similar looping.

webapp/src/main/java/rocks/metaldetector/service/artist/FollowArtistServiceImpl.java
```diff
@@ -41,13 +43,32 @@
...
+    );
+
+    if (!followActionRepository.existsByUserIdAndArtistId(currentUser.getId(), artistEntity.getId())) {
+      FollowActionEntity followAction = FollowActionEntity.builder()
+          .user(currentUser())
+          .artist(artistEntity)
+          .build();
+
+      followActionRepository.save(followAction);
+    }
```
DanielW1987, line 69: I think line 57 to 69 can be merged wirh the method above.

webapp/src/main/resources/static/ts/src/service/my-artists-render-service.ts
```diff
@@ -90,16 +99,26 @@ export class MyArtistsRenderService extends AbstractRenderService<MyArtistsRespo
...
         })
     }
 
-    private attachPagination(paginationData: Pagination) {
+    private attachPagination(paginationData: Pagination): void {
         const paginationList = this.paginationComponent.render(paginationData);
         this.hostElement.insertAdjacentElement("beforeend", paginationList);
     }
+
+    private renderImportButton(): void {
```
DanielW1987, line 115: I think the rendering is already done?! It's more an add of event listener and the method should be called within the constructor.

DanielW1987 (commented):
My comments for now. I have not looked at it in the frontend yet. But I will definitely do that on the weekend.

webapp/src/main/java/rocks/metaldetector/service/artist/ArtistServiceImpl.java
NilsDammenhayn: Yeah I thought about that at first. But then the Stream has a side effect on a list that is unrelated to this stream. That feels kind of weird. If the list that is given into this method is changed we could also make this method void and have the same effect in the calling class. Also not nice... To get rid of the second stream we need a `disjunct` function, guava has something like that. But just for this case I don't want to pull the dependency...
Long story short: I'd like to leave it that way :)

webapp/src/main/java/rocks/metaldetector/service/artist/FollowArtistServiceImpl.java
NilsDammenhayn: I just realized that I am not even using the new `follow()` method anymore, so it can go completely (and some other new stuff too...). I guess I wrote that in the beginning and didn't realize when it became obsolete. And I didn't test it... Very weird 🤷‍♂️

webapp/src/main/resources/static/ts/src/service/my-artists-render-service.ts
DanielW1987: You forgot to address this issue.

DanielW1987 (approved):

PR #220, opened 2020-10-28
Spotify token refreshing

hedgehog1833:
The `SpotifyUserAuthorizationService` now provides a method that returns the access token for the current user and refreshes it if necessary.


hedgehog1833:
[Use refresh token when user token has expired](https://trello.com/c/TpYZqTAi/239-use-refresh-token-when-user-token-has-expired)

DanielW1987 (approved):

PR #224, opened 2020-11-01
Reworked CurrentPublicUserIdSupplier to CurrentUserSupplier

hedgehog1833:
The `CurrentUserSupplier` returns the `UserEntity` of the current user.  One thing we have to be aware of now is, that this method potentially could return null. This is the case when we have an anonymous user. As long as we know that only logged in users are there this is fine. But if we for example would start to open the page up to anonymous users we would have to take care of the null-possibility.
What do you think?


hedgehog1833:
Oh and I have absolutely no clue why I could not use an import for the TestConfiguration-class in the integration test, it just didn't comile 🤷‍♂️

hedgehog1833:
[Service to retrieve the current user](https://trello.com/c/no9gJGrg/226-service-to-retrieve-the-current-user)

DanielW1987:
> One thing we have to be aware of now is, that this method potentially could return null. This is the case when we have an anonymous user.

I think we had the same problem before.

webapp/src/main/java/rocks/metaldetector/service/artist/FollowArtistServiceImpl.java
```diff
@@ -104,10 +104,6 @@ private ArtistEntity saveAndFetchArtist(String externalId, ArtistSource source)
     return artistRepository.save(artistEntity);
   }
 
-  private UserEntity currentUser() {
-    return fetchUserEntity(currentPublicUserIdSupplier.get());
-  }
-
```
DanielW1987, line 110: I think you forgot to remove `fetchUserEntity` in this class

webapp/src/test/java/rocks/metaldetector/security/CurrentUserSupplierImplIntegrationTest.java
```diff
@@ -0,0 +1,58 @@
...
+@SpringJUnitConfig
+class CurrentUserSupplierImplIntegrationTest implements WithAssertions, WithIntegrationTestConfig {
+
+  private static final String USERNAME = "user";
+  private static final String EMAIL = "user@mail.com";
+
+  private final CurrentUserSupplierImpl underTest = new CurrentUserSupplierImpl();
+
+  @Autowired
+  private UserDetailsService userDetailsService;
```
DanielW1987, line 24: you can remove that

webapp/src/test/java/rocks/metaldetector/security/CurrentUserSupplierImplIntegrationTest.java
```diff
@@ -0,0 +1,58 @@
...
+  void test_null_returned() {
+    // when
+    var result = underTest.get();
+
+    // then
+    assertThat(result).isNull();
+  }
+
+  @org.springframework.boot.test.context.TestConfiguration
+  static class TestConfiguration {
```
DanielW1987, line 51: it does not work, because name of annotation and class are equal. Rename the class and after that you can use `@TestConfiguration`

webapp/src/main/java/rocks/metaldetector/service/artist/FollowArtistServiceImpl.java
hedgehog1833: This method is still needed. It does not fetch the current, but any user for the given publicUserId. This is used by the NotificationService

DanielW1987: Check!

webapp/src/test/java/rocks/metaldetector/security/CurrentUserSupplierImplIntegrationTest.java
hedgehog1833: -.-^^

DanielW1987 (approved):

PR #225, opened 2020-11-03
Content-Security-Policy introduced

hedgehog1833:
With this PR a content security policy is introduced. All scripts come with a nonce value which is generated on the server per each request. This value is also sent as a header where it serves as an allowlist for scripts. All scripts coming with that value can be executed by the browser.
I added some sources to the trello card where the components of the csp are explained. The reason why I create the whole header in the filter (instead of using the possibility to configure it in the SecuritConfigs configure-method) is, that I was not able alter that header in my filter. The header has to contain the random nonce value so I have to format the string. I guess my filter ran before the SecurityConfig's filter, but I did not find out which filter does that or at which position in the filter chain it is set.
What's missing: it is possible to set a report url where csp reports are sent by the browser. We should configure that for production. I'll create a card for that.
Whats' not so nice: we have to remember a few things from now on: all new frontend pages have to be added to the list of pages so that the filter accesses them. Also the script blocks have to have the nonce-tag.


hedgehog1833:
[Content security policy (CSP)](https://trello.com/c/F6DAAUBR/224-content-security-policy-csp)

hedgehog1833:
@DanielW1987 I added everything we talked about. There are no more error messages in the browser log. If needed logging can be activated via a env property.

webapp/src/main/java/rocks/metaldetector/security/CspNonceFilter.java
```diff
@@ -16,7 +16,7 @@
 public class CspNonceFilter implements Filter {
 
   static final String CSP_HEADER_NAME = "Content-Security-Policy";
-  static final String CSP_POLICY = "object-src 'none'; script-src 'nonce-script-%s' 'strict-dynamic'; base-uri 'none';";
+  static final String CSP_POLICY = "object-src 'none'; script-src 'nonce-%s' 'strict-dynamic'; style-src https://fonts.googleapis.com 'nonce-%s'; base-uri 'none'; report-uri /rest/v1/csp-violation-report; report-to /rest/v1/csp-violation-report;";
```
DanielW1987, line 19: maybe you can add the static Endpoint constant here

webapp/src/main/java/rocks/metaldetector/security/CspNonceFilter.java
```diff
@@ -25,7 +25,7 @@
   public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
     String nonce = nonceSupplier.get();
     servletRequest.setAttribute(ATTRIBUTE_NAME, nonce);
-    ((HttpServletResponse) servletResponse).setHeader(CSP_HEADER_NAME, CSP_POLICY.formatted(nonce));
+    ((HttpServletResponse) servletResponse).setHeader(CSP_HEADER_NAME, CSP_POLICY.formatted(nonce, nonce));
```
DanielW1987, line 28: don't understand why you pass the nonce a second time into `formatted` method

DanielW1987 (commented):
The tests are failing.

webapp/src/main/java/rocks/metaldetector/security/CspNonceFilter.java
hedgehog1833: Because the nonce is also used for the `style-src`. Two `%s` in the String and two parameters to the format method.

DanielW1987 (approved):

PR #228, opened 2020-11-07
Endpoint to follow multiple spotify artists

hedgehog1833:
A new endpoint to follow multiple spotify artists. This endpoint fetched all artists from Spotify that are not yet in our database and then saves all FollowActionEntities for the current user.

### DIFF
--- a/webapp/src/main/java/rocks/metaldetector/service/artist/FollowArtistServiceImpl.java
+++ b/webapp/src/main/java/rocks/metaldetector/service/artist/FollowArtistServiceImpl.java
@@ -33,6 +33,7 @@ public class FollowArtistServiceImpl implements FollowArtistService {
   private final DiscogsService discogsService;
   private final ArtistTransformer artistTransformer;
   private final CurrentPublicUserIdSupplier currentPublicUserIdSupplier;
+  private final ArtistService artistService;
 
   @Override
   @Transactional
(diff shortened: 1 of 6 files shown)


hedgehog1833:
[Artist import endpoints](https://trello.com/c/ILFpbOQk/243-artist-import-endpoints)

webapp/src/main/java/rocks/metaldetector/service/artist/FollowArtistServiceImpl.java
```diff
@@ -104,6 +120,13 @@ private ArtistEntity saveAndFetchArtist(String externalId, ArtistSource source)
     return artistRepository.save(artistEntity);
   }
 
+  private List<ArtistEntity> saveAndFetchArtists(List<String> spotifyArtistIds) {
+    List<String> newArtistsIds = artistService.findNewArtistIds(spotifyArtistIds);
+    List<SpotifyArtistDto> spotifyArtistDtos = spotifyService.searchArtistsByIds(newArtistsIds);
+    artistService.persistArtists(spotifyArtistDtos);
+    return artistRepository.findAllByExternalIdIn(spotifyArtistIds);
```
DanielW1987, line 127: This is somewhat redundant.

Via `artistService.persistArtists(spotifyArtistDtos)` the the method `artistRepository.saveAll(artistEntities)` is called and this already returns all ArtistEntities.

hedgehog1833: No :) Only those artists are persisted, that we don't already have in our database. But what we need are all artists that were requested. So we need this call here.

DanielW1987: maybe you can name `spotifyArtistDtos` to `newSpotifyArtistDtos` to make it less confusing.

DanielW1987 (approved):

PR #245, opened 2020-11-29
Finish spotify synchro

DanielW1987:
I adapted the backend part and also cleaned up a bit with the naming.

Unfortunately 2 files are also marked as deleted and newly created. This is due to renaming.

Now it is only open that you can also synchronize your saved artists from Spotify. But for this I make a separate card.

Pagination is not so easy at this point, because we would have to remember the selected artists across pages. Let's first look without Pagination to see if this is practicable.


DanielW1987:
[Frontend to synchronize artists from Spotify](https://trello.com/c/q4gIirGx/247-frontend-to-synchronize-artists-from-spotify)

hedgehog1833 (changes requested):
As discussed, this looks pretty fine except for this naming thing, which pulls through almost every file 😅

hedgehog1833 (approved):
👍

PR #246, opened 2020-11-30
Reduce javascript file size

DanielW1987:
Since we use `moment.js`, the config for all locales (which are mostly translations) is generated in the bundles.
This makes the bundles very large (>300 KB).

This can be easily configured via Webpack. Nevertheless I decided to use a different lib because it is much smaller and has a similar feature set.

The new sizes in prod are as follows:
```
asset my_artists.bundle.js 40.9 KiB [emitted] [minimized] (name: my_artists)
asset releases.bundle.js 40.1 KiB [emitted] [minimized] (name: releases)
asset homepage.bundle.js 38 KiB [emitted] [minimized] (name: homepage)
asset search.bundle.js 34.7 KiB [emitted] [minimized] (name: search)
asset spotify_synchronization.bundle.js 30.8 KiB [emitted] [minimized] (name: spotify_synchronization)
```

### DIFF
--- a/webapp/src/main/resources/static/ts/package.json
+++ b/webapp/src/main/resources/static/ts/package.json
@@ -23,13 +23,12 @@
     "ts-loader": "^8.0.11",
     "typescript": "^4.1.2",
     "webpack": "^5.9.0",
-    "webpack-cli": "^4.2.0",
-    "webpack-dev-server": "^3.11.0"
+    "webpack-cli": "^4.2.0"
   },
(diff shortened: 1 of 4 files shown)


DanielW1987:
[Huge javascript file size after using moment.js](https://trello.com/c/EFcGbEqk/246-huge-javascript-file-size-after-using-momentjs)

hedgehog1833 (approved):

PR #247, opened 2020-11-30
handle long artist names

DanielW1987:
A line break no longer leads to format shifts


DanielW1987:
[Handle long artist names on startpage and My artists page](https://trello.com/c/NC6e34Ml/245-handle-long-artist-names-on-startpage-and-my-artists-page)

hedgehog1833 (approved):

PR #255, opened 2020-12-03
use local images

DanielW1987:
Everything worked fine for me locally after I cleaned up my docker environment a bit. Concretely I have done the following:

- Stopped all running containers
- All containers removed
- All volumes removed
- `docker system prune`
- `docker-compose pull`

Then start the docker containers as normal.


DanielW1987:
[Images must be saved in the Default Profile so that they can be displayed.](https://trello.com/c/xJsONlqC/241-images-must-be-saved-in-the-default-profile-so-that-they-can-be-displayed)

hedgehog1833 (approved):

PR #257, opened 2020-12-05
Sorting on releases page by release date or announcement date

hedgehog1833:
I tried out some stuff for sorting by release date or announcement date.

### DIFF
--- a/webapp/src/main/resources/static/ts/src/clients/releases-rest-client.ts
+++ b/webapp/src/main/resources/static/ts/src/clients/releases-rest-client.ts
@@ -26,6 +26,8 @@ export class ReleasesRestClient {
     }
 
     private fetchReleases(url: string): Promise<ReleasesResponse> {
+        const parameters = this.urlService.getParametersFromUrl("sort");
+        const sortParameter = parameters.length > 0 ? "?sort=" + parameters[0] + "&sort=" + parameters[1] + "&sort=" + parameters[2] : "";
         axiosConfig.params = {
             page: this.urlService.getPageFromUrl(),
             size: 30,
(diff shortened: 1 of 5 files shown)


hedgehog1833:
[Releases Page: Sorting for announcement date](https://trello.com/c/8FWlzuVp/238-releases-page-sorting-for-announcement-date)

webapp/src/main/resources/static/ts/src/clients/releases-rest-client.ts
```diff
@@ -26,14 +26,16 @@ export class ReleasesRestClient {
...
+        const sortParameter = parameters.length > 0 ? "?sort=" + parameters[0] + "&sort=" + parameters[1] + "&sort=" + parameters[2] : "";
         axiosConfig.params = {
             page: this.urlService.getPageFromUrl(),
             size: 30,
             dateFrom: this.dateFormatService.format(new Date().toUTCString(), DateFormat.UTC)
         }
 
         return axios.get(
-            url,
+            url + sortParameter,
```
hedgehog1833, line 38: I did not use the axiosConfig.params because I have multiple sort parameter. If I just pass the string with all three to `params` the string gets url encoded whitch leads to weird behaviors in the rest endpoint. So just passing it as part of the URL seemed the easiest way :)

hedgehog1833: Of course passing a list works... 😄

DanielW1987 (approved):
As a first version it fits.

Ask me the following question:
- Why does the URL need such a long parameter. At the moment it's only about sorting by Release Date or Created Date. Sorting by artist and album should be implicitly set by the backend.
- I'm not sure yet if I want to have the sorting in the **filter** bar.

Sorry, at the frontend I become a Rolf.

hedgehog1833:
Haha no worries :D Yeah I'll merge it and we can check on the rest later!

PR #260, opened 2020-12-06
codecov.yml introduced

MetalDetectorRocks:


hedgehog1833:
Works \o/ Would have worked with my own user too, the problem was that full path names instead of packages have to be put into the yml. I just took the default yml and added the ignores.
config packages could also be ignored or what do you think?

DanielW1987:
I would like to leave the Conigs in it. At least in the integration tests, which don't have their own config, they should be tested.

DanielW1987 (approved):

PR #261, opened 2020-12-14
show info message if there are no artists from spotify to synchronize

DanielW1987:


DanielW1987:
[Show info message if no artists could be fetched from spotify](https://trello.com/c/seeytTeQ/257-show-info-message-if-no-artists-could-be-fetched-from-spotify)

hedgehog1833 (approved):

PR #262, opened 2020-12-14
Filter 'All upcoming' should also show releases from today

DanielW1987:


DanielW1987:
[Release Page: Filter 'All upcoming' should also show releases from today](https://trello.com/c/JdvOVngv/265-release-page-filter-all-upcoming-should-also-show-releases-from-today)

hedgehog1833 (commented):
I'd like the member to be named like the class, so dateService instead of dateFormatService. In Java/Groovy this should have happened automatically 🤔

hedgehog1833 (approved):

PR #264, opened 2020-12-16
preparation for docker swarm

DanielW1987:
Docker Swarm manages the secrets encrypted on the host. If a container shall use a secret, docker mounts it via a file in the container. The container must therefore be smart enough to set an ENV from the contents of a file. This is now done by the new `docker-entrypoint.sh` that is copied into the docker image. 

For example, `docker run` can be executed as follows:

```bash
docker run -e JWT_SECRET=my-secret metaldetector/metal-detector
```

We already know that.

Now, also possible is:

```bash
docker run -e JWT_SECRET_FILE=/path/to/secret metaldetector/metal-detector
```

The last variant will expose the contents of the file to the ENV `JWT_SECRET`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,5 +27,7 @@ LABEL org.label-schema.vcs-url="https://github.com/MetalDetectorRocks/metal-dete
 LABEL org.label-schema.vcs-ref=$VCS_REF
 
 COPY $SOURCE_JAR_FILE app.jar
+COPY docker-entrypoint.sh /app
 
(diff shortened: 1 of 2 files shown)


DanielW1987:
[Prod Deployment via Hetzner Cloud](https://trello.com/c/eG6fmUga/244-prod-deployment-via-hetzner-cloud)

hedgehog1833 (approved):
lgtm

PR #270, opened 2020-12-19
Notification Config Backend

hedgehog1833:
Users have a default notification config which can be updated via rest endpoint. I started with the NotificationEntity having the one-to-one to the UserEntity but this way it is not that easy having a default automatically created when the user is created so I switched it to the User (making is bigger again, i know...).
Second maybe a bit weird point is that the endpoint to update the config is `/rest/v1/users/notification-config` - without the user id after `users`. So it is only possible to change one's own config currently. This way we don't have to worry some calling the endpoint with userIds other than his own. What's your opinion on this?

### DIFF
--- a/persistence/src/main/java/rocks/metaldetector/persistence/domain/notification/NotificationConfigEntity.java
+++ b/persistence/src/main/java/rocks/metaldetector/persistence/domain/notification/NotificationConfigEntity.java
@@ -1,0 +1,41 @@
+package rocks.metaldetector.persistence.domain.notification;
+
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.NonNull;
+import lombok.Setter;
+import rocks.metaldetector.persistence.domain.BaseEntity;
+import rocks.metaldetector.persistence.domain.user.UserEntity;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.OneToOne;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor(access = AccessLevel.PRIVATE) // for lombok builder
+@Builder
+@EqualsAndHashCode(callSuper = true)
+@Entity(name = "notificationConfigs")
+public class NotificationConfigEntity extends BaseEntity {
+
+  @OneToOne(mappedBy = "notificationConfig")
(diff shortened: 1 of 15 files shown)


hedgehog1833:
[Make Notifications configurable](https://trello.com/c/hlDkMGvu/255-make-notifications-configurable)

persistence/src/main/java/rocks/metaldetector/persistence/domain/notification/NotificationConfigEntity.java
```diff
@@ -0,0 +1,41 @@
...
+@EqualsAndHashCode(callSuper = true)
+@Entity(name = "notificationConfigs")
+public class NotificationConfigEntity extends BaseEntity {
+
+  @OneToOne(mappedBy = "notificationConfig")
+  private UserEntity user;
+
+  @Column(name = "frequency", columnDefinition = "integer default 4")
+  @NonNull
+  private Integer frequency = 4;
```
DanielWagner87, line 32: This naming is not accurate enough. 4 Times per Day, per Month? Maybe something like `frequencyInWeeks`?

persistence/src/main/java/rocks/metaldetector/persistence/domain/user/UserEntity.java
```diff
@@ -68,6 +72,10 @@
   @Column(name = "last_login")
   private LocalDateTime lastLogin;
 
+  @OneToOne(cascade = CascadeType.ALL)
+  @JoinColumn(name = "notification_config", referencedColumnName = "id")
+  private NotificationConfigEntity notificationConfig;
```
DanielWagner87, line 77: Not sure we really need that here. As soon as a user is created, a NotificationConfig with default values is created at the same time. But this would have to be done explicitly once. 

And of course we also need a `NotificationConfigRepository` with `getNotificationConfigByUserId()`.

DanielWagner87:
> Second maybe a bit weird point is that the endpoint to update the config is /rest/v1/users/notification-config - without the user id after users

Does not have to be a new endpoint. We have an endpoint to update the user. The config belongs to a user. We should think more in the direction of REST. But it can also be complicated to find out whether the email, the password, the user picture or the config should be changed. Therefore I would leave it as it is.

In addition, we also need something to activate or deactivate the notification at all. You should have the possibility to turn off annoying emails. That is missing here. Can also be shipped with another PR, but it needs to be persisted, so it fits into this PR

hedgehog1833:
I kind of destroyed the branch, I'm gonna close and reopen this PR.

PR #276, opened 2020-12-22
Metal Detector gradle migration

hedgehog1833:
I migrated the Detector to gradle. Some DTOs needed a no-args constructor and I had to add an annotation somewhere but all in all it was pretty straight forward. Build time dropped from about 1:50 to less than 1:20.


hedgehog1833:
Info for the local setup after merging this: uncheck the "Generate *.iml files...." option in your Idea Gradle settings. We don't need those.

.github/workflows/deploy.yml
```diff
@@ -29,8 +29,11 @@ jobs:
         with:
           java-version: 1.15
 
-      - name: Build with Maven
-        run: mvn -B clean verify --file pom.xml
+      - name: Execute all tests
+        uses: eskatos/gradle-command-action@v1
+        with:
+          gradle-version: 6.7.1
+          arguments: clean test
```
DanielW1987, line 36: have you checked if this creates a .jar?

build.gradle
```diff
@@ -0,0 +1,94 @@
...
+    }
+    compileOnly {
+      extendsFrom annotationProcessor
+    }
+  }
+
+  dependencies {
+    implementation "org.springframework.boot:spring-boot-starter-security:$springBootVersion"
+    implementation "org.springframework.boot:spring-boot-starter-cache:$springBootVersion"
+    implementation "org.springframework.boot:spring-boot-configuration-processor:$springBootVersion"
```
DanielW1987, line 56: do we really have to use a concrete springBootVersion? With Maven, this is managed by the parent.

persistence/src/main/java/rocks/metaldetector/persistence/domain/token/TokenRepository.java
```diff
@@ -10,9 +11,9 @@
 public interface TokenRepository extends JpaRepository<TokenEntity, Long> {
 
   @Query(value = "SELECT * FROM tokens WHERE TOKEN_TYPE = 'EMAIL_VERIFICATION' AND TOKEN_STRING = :tokenString", nativeQuery = true)
-  Optional<TokenEntity> findEmailVerificationToken(String tokenString);
+  Optional<TokenEntity> findEmailVerificationToken(@Param("tokenString") String tokenString);
 
   @Query(value = "SELECT * FROM tokens WHERE TOKEN_TYPE = 'PASSWORD_RESET' AND TOKEN_STRING = :tokenString", nativeQuery = true)
-  Optional<TokenEntity> findResetPasswordToken(String tokenString);
+  Optional<TokenEntity> findResetPasswordToken(@Param("tokenString") String tokenString);
```
DanielW1987, line 17: why do we need this?

DanielW1987 (commented):
I miss the CodeCov post on this PR

persistence/src/main/java/rocks/metaldetector/persistence/domain/token/TokenRepository.java
hedgehog1833: Because otherwise the corresponding integration test fails 🤷‍♂️  The annotation explicitly says that this parameter is exchanged with the string in the query.

build.gradle
hedgehog1833: I think so. In gradle there is no such thing as a parent so we have to pull the version by ourselves. But it doesn't make much difference where to write the version number, so who cares? 😉

DanielW1987: In Butler we don't have to specify the versions explicitly either.

hedgehog1833: Hmm that's right, it's just in the plugin there but I can't get it running without the version number here.... 🤔 I can check this again, but I'd say that's a point for later 😉

DanielW1987: Maybe you just have to remove the `apply false`. Why is this there?

hedgehog1833: I tried that, doesn't work.

DanielW1987 (approved):

build.gradle
```diff
@@ -0,0 +1,127 @@
...
+    apacheCommonsTextVersion = "1.9"
+    h2Version = "1.4.200"
+    jacksonVersion = "2.12.0"
+    jacocoVersion = "0.8.6"
+    jaxbApiVersion = "2.3.1"
+    junitVersion = "5.7.0"
+    mockitoVersion = "3.6.28"
+    springBootVersion = "2.3.7.RELEASE"
+    lombokVersion = "1.18.16"
+    springSecurityVersion = "5.2.6.RELEASE"
```
DanielW1987, line 12: can you please define all versions here? Now we have versions in many different places.

DanielW1987 (changes requested):
set java version also in `system.properties`

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:15-slim-buster
+FROM openjdk:11-stretch
```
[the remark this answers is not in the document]
hedgehog1833: Will do, but we use this one in the butler so we should change it there too.

DanielW1987 (approved):

PR #279, opened 2020-12-28
fix docker health check

DanielW1987:
We can't use the Docker healthchecks on Heroku, so we'll deploy the .jar there again.


hedgehog1833 (approved):
lgtm

PR #283, opened 2020-12-30
email layout

DanielW1987:
I had to set the link to a few resources statically, so we won't see the final result until it's deployed.


DanielW1987:
[Uniform layout for the e-mails](https://trello.com/c/LTLXkOVW/12-uniform-layout-for-the-e-mails)

hedgehog1833 (approved):
👌

PR #285, opened 2020-12-30
fix spotify synchro issue

DanielW1987:
The API for querying all followed artists of a user does not work with the offset parameter, but uses a different mechanism. See <https://developer.spotify.com/documentation/web-api/reference/follow/get-followed/>


DanielW1987:
[Spotify does not sync all bands](https://trello.com/c/TZZLpG8C/281-spotify-does-not-sync-all-bands)

hedgehog1833 (approved):

PR #289, opened 2021-01-06
Spotfiy search for queries with Umlaut fixed

hedgehog1833:
The url-encoding is done by the rest template. If we do this too, it breaks -.-


hedgehog1833:
[Spotify doesn't work if artist name has an umlaut or special character](https://trello.com/c/Q34KHyXp/263-spotify-doesnt-work-if-artist-name-has-an-umlaut-or-special-character)

DanielW1987 (approved):

PR #290, opened 2021-01-06
Registration code is required to create a new account

hedgehog1833:
Nothing fancy, just a check for a code that has to be present as an environment variable on account creation via the html form.


hedgehog1833:
[Registration with code](https://trello.com/c/tiLF3qQM/288-registration-with-code)

webapp/src/main/java/rocks/metaldetector/web/controller/mvc/authentication/RegistrationController.java
```diff
@@ -54,7 +56,8 @@ public ModelAndView showRegistrationForm() {
   @PostMapping(Endpoints.Guest.REGISTER)
   public ModelAndView registerUserAccount(@Valid @ModelAttribute RegisterUserRequest registerUserRequest, BindingResult bindingResult) {
     // show registration form if there are validation errors
-    if (bindingResult.hasErrors()) {
+    String registrationCode = environment.getProperty("REGISTRATION_CODE");
+    if (bindingResult.hasErrors() || !registerUserRequest.getRegistrationCode().equals(registrationCode)) {
```
DanielW1987, line 60: maybe equalsIgnoreCase here

DanielW1987 (commented):
what about the registration enabling? The registration is disabled on prod.

DanielW1987 (approved):

PR #293, opened 2021-01-07
Notification is sent based on frequency configuration

hedgehog1833:
The `NotificationService` now iterates over all NotificationConfigs instead of all users. Based on if notifcations are enabled, the user is active and the configured frequency and date of the last notification.
I removed the endpoint for notifying a certain user, because we can't get NotificationConfigEntity based on the publicUserId.


hedgehog1833:
[Make Notifications configurable](https://trello.com/c/hlDkMGvu/255-make-notifications-configurable)

hedgehog1833:
@DanielW1987 I refactored the things we discussed. The Butler is now called twice for every supported frequency (currently 2 and 4 weeks) instead of twice for every user, but with all artists. I built the tests in a way that more frequencies can be added without having to adjust tests. Also did some cleanups in the service's test class.
This can be reviewed now.

webapp/src/main/java/rocks/metaldetector/service/email/ReleasesEmail.java
```diff
@@ -5,18 +5,20 @@
 
 import java.util.List;
 
-public final class NewReleasesEmail extends AbstractEmail {
+public final class ReleasesEmail extends AbstractEmail {
 
-  public static final String SUBJECT = "Your new releases";
+  public static final String SUBJECT = "Your releases";
```
DanielW1987, line 10: Maybe something like "Your latest release updates"?

webapp/src/main/java/rocks/metaldetector/service/notification/NotificationServiceImpl.java
```diff
@@ -35,27 +39,18 @@
...
-      List<ReleaseDto> newReleases = releaseService.findAllReleases(followedArtistsNames, new TimeRange(now, now.plusMonths(3)));
-
-      if (!newReleases.isEmpty()) {
-        emailService.sendEmail(new NewReleasesEmail(user.getEmail(), user.getUsername(), newReleases));
-      }
-    }
+    notificationConfigRepository.findAll().stream()
+        .filter(config -> config.getUser().isEnabled() &&
+                          config.getNotify())
+        .forEach(notificationConfig -> frequencyNotification(notificationConfig, getReleaseContainerForConfig(releaseContainer, notificationConfig)));
```
DanielW1987, line 49: I'm confused aboud `config` vs. `notificationConfig` (see `filter` lambda above)

```suggestion
        .forEach(config -> frequencyNotification(config, getReleaseContainerForConfig(releaseContainer, config)));
```

webapp/src/main/java/rocks/metaldetector/service/notification/NotificationServiceImpl.java
```diff
@@ -76,4 +72,56 @@ public void updateCurrentUserNotificationConfig(NotificationConfigDto notificati
...
+          emailService.sendEmail(new ReleasesEmail(user.getEmail(), user.getUsername(), upcomingReleases, recentReleases));
+
+          notificationConfigEntity.setLastNotificationDate(now);
+          notificationConfigRepository.save(notificationConfigEntity);
+        }
+      }
+    }
+  }
+
+  private List<ReleaseContainer> createReleaseContainer() {
```
DanielW1987, line 100: Maybe this should return a Map:

```
{
  2: ReleaseContainer for frequency 2,
  4: ReleaseContainer for frequency 4,
}
```

Then you don't need method `getReleaseContainerForConfig`.

webapp/src/main/resources/templates/email/releases-email.html
```diff
@@ -23,7 +23,26 @@
...
+    <table class="table">
+        <thead>
+        <tr>
+            <th scope="col">Artist</th>
+            <th scope="col">Album Title</th>
+            <th scope="col">Release Date</th>
+        </tr>
+        </thead>
+        <tbody>
+        <tr th:each="release: ${recentReleases}">
```
DanielW1987, line 45: We have a problem here if one of the lists is empty, e. g. 2 upcoming releases and 0 recrent releases. Good enough for now. We should take a look at it later -> Trello card.

DanielW1987 (approved):

PR #300, opened 2021-01-16
Updating the current user's email address now working

hedgehog1833:
Currently the whole current user dto is fetched on get and update although we only need the email. Do you think this is an issue (security-wise for example?).  Of course the dto does not contain a password.
Another thing: Do you think it is an issue that after this PR updating the email address will work but the banner is still there? I could also disable the button or something..

### DIFF
--- a/webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
+++ b/webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
@@ -3,8 +3,6 @@ package rocks.metaldetector.service.user;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.codec.digest.DigestUtils;
-import org.springframework.data.domain.PageRequest;
-import org.springframework.data.domain.Pageable;
 import org.springframework.security.authentication.LockedException;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
@@ -128,6 +126,10 @@ public class UserServiceImpl implements UserService {
   @Override
   @Transactional
   public UserDto updateCurrentEmail(String emailAddress) {
+    if (userRepository.existsByEmail(emailAddress)) {
+      throw new IllegalArgumentException("emailAddress already in use");
+    }
+
     UserEntity currentUser = currentUserSupplier.get();
     currentUser.setEmail(emailAddress);
     UserEntity updatedUser = userRepository.save(currentUser);
@@ -168,13 +170,9 @@ public class UserServiceImpl implements UserService {
 
   @Override
   @Transactional(readOnly = true)
-  public List<UserDto> getAllUsers(int page, int limit) {
-    Pageable pageable = PageRequest.of(page, limit);
-
-    return userRepository.findAll(pageable)
-        .stream()
-        .map(userTransformer::transform)
-        .collect(Collectors.toList());
+  public UserDto getCurrentUser() {
+    UserEntity currentUser = currentUserSupplier.get();
+    return userTransformer.transform(currentUser);
   }
 
   @Override
--- a/webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
+++ b/webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
@@ -127,6 +127,15 @@ public class UserServiceImpl implements UserService {
 
   @Override
   @Transactional
+  public UserDto updateCurrentEmail(String emailAddress) {
+    UserEntity currentUser = currentUserSupplier.get();
+    currentUser.setEmail(emailAddress);
+    UserEntity updatedUser = userRepository.save(currentUser);
+    return userTransformer.transform(updatedUser);
+  }
+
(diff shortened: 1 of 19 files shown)


hedgehog1833:
[Change email address](https://trello.com/c/UOp78tIZ/258-change-email-address)

webapp/src/main/java/rocks/metaldetector/web/controller/rest/UserRestController.java
```diff
@@ -53,6 +59,15 @@
...
 
+  @GetMapping(path = CURRENT,
+              produces = APPLICATION_JSON_VALUE)
+  public ResponseEntity<UserResponse> getCurrentUser() {
+    UserEntity currentUser = currentUserSupplier.get();
+    UserResponse response = mapper.map(currentUser, UserResponse.class);
+
+    return ResponseEntity.ok(response);
+  }
+
```
DanielW1987, line 70: Due to the "new" endpoint design, the `current` methods should be in their own controller. Something like `UserAccountRestController` for everything the user can do with his own account. The `UserRestController` contains methods that the admin can do with one or with a set of users.

DanielW1987: Can we create or do we already have a service method for

```java
UserEntity currentUser = currentUserSupplier.get();
UserResponse response = mapper.map(currentUser, UserResponse.class);
```

As a convention we do not want tu use any persistent types in cotroller.

webapp/src/main/resources/static/ts/src/service/account-details-render-service.ts
```diff
@@ -0,0 +1,33 @@
...
+        const updateEmailAddressButton = document.getElementById("update-email-address") as HTMLButtonElement;
+        updateEmailAddressButton.addEventListener("click", this.onUpdateEmailAddressClicked.bind(this));
+    }
+
+    private onUpdateEmailAddressClicked(): void {
+        const inputElement = document.getElementById("email-address")! as HTMLInputElement;
+        this.accountDetailsRestClient.updateEmailAddress(inputElement.value)
+          .then(response => inputElement.value = response.email)
+          .then(() => this.toastService.createInfoToast("Successfully updated email address!"));
+    }
```
DanielW1987, line 32: We should not use `ToastService` everywhere where we want to give feedback to the user. 

It's nice for following and unfollowing, but not absolutely necessary, because you can see via the icon if it worked or not.

The problem is that the Toast concept does not work great on mobile devices. It can stay that way for now, but we should use the toasts more consciously in the future. A green checkmark next to the e-mail field or something like that would probably be better here.

DanielW1987: ...and Toasts in Admin area is fine at every place, because admin area is only for us.

webapp/src/test/java/rocks/metaldetector/testutil/BaseWebMvcTestWithSecurity.java
```diff
@@ -49,4 +50,7 @@
   @MockBean
   protected HttpSecurity httpSecurity;
 
+  @MockBean
+  protected CurrentUserSupplier currentUserSupplier;
+
```
DanielW1987, line 55: You can remove that if you move the current user logic into the service.

webapp/src/main/java/rocks/metaldetector/web/controller/rest/UserRestController.java
```diff
@@ -74,4 +89,14 @@
 
     return ResponseEntity.status(HttpStatus.OK).body(response);
   }
+
+  @PutMapping(path = CURRENT + EMAIL,
+              consumes = APPLICATION_JSON_VALUE,
+              produces = APPLICATION_JSON_VALUE)
+  public ResponseEntity<UserResponse> updateCurrentEmail(@Valid @RequestBody UpdateEmailRequest request) {
```
DanielW1987, line 96: We should only return the new email address here.

DanielW1987:
> Do you think it is an issue that after this PR updating the email address will work but the banner is still there? I could also disable the button or something.

I think we can leave it as it is...

webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
```diff
@@ -125,6 +125,15 @@ public UserDto updateUser(String publicId, UserDto userDto) {
...
   }
 
+  @Override
+  @Transactional
+  public UserDto updateCurrentEmail(String emailAddress) {
+    UserEntity currentUser = currentUserSupplier.get();
+    currentUser.setEmail(emailAddress);
+    UserEntity updatedUser = userRepository.save(currentUser);
+    return userTransformer.transform(updatedUser);
+  }
```
DanielW1987, line 135: What happens if I update with an email that already exists. There is a UNIQUE constraint on the email field in the database.

DanielW1987 (commented):
At a later point, we should still consider whether the workflow is sufficient as it is. 

I could also specify an email here that does not belong to me. 

But this should then be implemented with another card.

DanielW1987 (approved):
I have created a Trello card with my findings: https://trello.com/c/C0YejgKL

PR #307, opened 2021-01-24
Users can delete their accounts

hedgehog1833:
Users can delete their accounts. On click on the button a popup asks for confirmation and only then the service is called. The UserService publishes an event which handles all the actions to take (clearing repositories and tables, sending an email) and then clears the security context and invalidates the user's session. The redirect is done in the frontend.


hedgehog1833:
[Delete Account](https://trello.com/c/pTPZfbUr/260-delete-account)

webapp/src/main/java/rocks/metaldetector/service/user/events/UserDeletionEventListener.java
```diff
@@ -0,0 +1,63 @@
...
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+  private final EmailService emailService;
+
+  @Override
+  @Transactional
+  public void onApplicationEvent(UserDeletionEvent event) {
+    UserEntity user = event.getUserEntity();
+    log.info("User '" + user.getPublicId() + "' deleted");
+
+    Optional<NotificationConfigEntity> notificationConfigOptional = notificationConfigRepository.findByUserId(user.getId());
```
DanielW1987, line 46: Maybe a specific deleteByUserId make sense for notification and spotify repository

webapp/src/main/java/rocks/metaldetector/service/user/events/UserDeletionEventPublisher.java
```diff
@@ -0,0 +1,18 @@
+package rocks.metaldetector.service.user.events;
+
+import lombok.AllArgsConstructor;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.stereotype.Component;
+import rocks.metaldetector.persistence.domain.user.UserEntity;
+
+@Component
+@AllArgsConstructor
+public class UserDeletionEventPublisher {
```
DanielW1987, line 10: Do we realy need an own publisher?

webapp/src/main/resources/static/ts/src/clients/account-details-rest-client.ts
```diff
@@ -40,4 +40,15 @@ export class AccountDetailsRestClient {
...
     }
+
+    public async deleteAccount(): Promise<void> {
+        return await axios
+            .delete(this.CURRENT_USER_ENDPOINT, axiosConfig)
+            .then(() => {
+                return;
+            })
+            .catch((error: AxiosError) => {
+                throw error;
```
DanielW1987, line 51: we should have a console.error() here.

webapp/src/main/resources/templates/frontend/account-details.html
```diff
@@ -45,8 +45,30 @@ <h2 class="h5 mt-5">Change password</h2>
...
+                            <div class="modal-header">
+                                <h5 class="modal-title">Delete Account</h5>
+                                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                                    <span aria-hidden="true">&times;</span>
+                                </button>
+                            </div>
+                            <div class="modal-body">
+                                <div class="small" role="alert">
+                                    <p>Dou you really want to delete your account? All data related to your account will be lost.</p>
+                                </div>
```
DanielW1987, line 63: Can you increase the font size a bit? Find that very small.

webapp/src/main/java/rocks/metaldetector/service/user/events/UserDeletionEventListener.java
hedgehog1833: Hm yeah for the NotificationConfig yes, because it is always present. But the SpotifyAuthEntity might not exist, so we had to check for its existance anyway right?

webapp/src/main/java/rocks/metaldetector/service/user/events/UserDeletionEventPublisher.java
hedgehog1833: 💡 we don't, I can just inject the publisher in the UserService

DanielW1987: dancing pikachu

webapp/src/main/java/rocks/metaldetector/service/user/events/UserDeletionEventListener.java
DanielW1987: > But the SpotifyAuthEntity might not exist, so we had to check for its existance anyway right?
Do not think you have to

webapp/src/main/java/rocks/metaldetector/service/user/events/UserDeletionEventPublisher.java
hedgehog1833: I think that's a custom emoji from slack 😉

DanielW1987: i know...please imagine it

DanielW1987 (approved):

PR #319, opened 2021-01-24
Fix pagination when filtering + Rework of sorting + display announcement date

DanielW1987:
3 things I have implemented here:

- The Pagination Component now always takes the current URL parameters into account, so that nothing can get lost. 
- The sorting in the Detector frontend is no longer dependent on how the sorting in the Butler works. This makes the code in the Detector frontend and backend simpler and easier to understand.
- The Announcement Date is now shown when sorting by it (but only then)


hedgehog1833 (approved):
❤️

DanielW1987:
[Pagination with Filter and Sorting broken](https://trello.com/c/Hvnv2Do9/266-pagination-with-filter-and-sorting-broken)

PR #321, opened 2021-01-24
Dependabot ignores restAssured and jakarta.xml.bind-api

hedgehog1833:
restAssured 4.3 has a bug and does not work with spring-boot 2.4 and all our dependencies use the bind-api in 2.x.x so 3 does not work here.
This should close the existing dependabot PRs and open one new for restAssured 4.2.1


DanielW1987 (approved):

hedgehog1833:
[DependaBot](https://trello.com/c/9HROCNKC/286-dependabot)

PR #325, opened 2021-01-24
Dependabot ignore removed

hedgehog1833:
First the ignore did not work as I expected and second I found out that the default behavior if you close a PR from the bot is that it won't notify about this update again but only about newer versions. That is exactly what I want here so I'll just remove the ignore part again.
Also I manually bumped (I like the word 😄) restAssured to 4.2.1, the currently latest functioning version.


DanielW1987 (approved):

hedgehog1833:
[DependaBot](https://trello.com/c/9HROCNKC/286-dependabot)

PR #335, opened 2021-02-06
email validation

DanielW1987:
I have now decided against separate client-side validation. The library I wanted to try out needs jQuery. We currently use only a very small part of jQuery and perspectively I want to throw it out. So makes no sense to bring a new lib in that needs jQuery.

Angular (or React) again have their own mechanisms to do form validation.

So for the few forms we have, we can leave it at pure server-side validation.

At is now also possible to submit the email with ENTER. It would be good if you could do the same with the password.


DanielW1987:
[Email validation and success message](https://trello.com/c/C0YejgKL/302-email-validation-and-success-message)

webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
```diff
@@ -131,11 +131,11 @@ public UserDto updateUser(String publicId, UserDto userDto) {
   @Override
   @Transactional
   public UserDto updateCurrentEmail(String emailAddress) {
-    if (userRepository.existsByEmail(emailAddress)) {
-      throw new IllegalArgumentException("emailAddress already in use");
+    UserEntity currentUser = currentUserSupplier.get();
+    if (!currentUser.getEmail().equalsIgnoreCase(emailAddress) && userRepository.existsByEmail(emailAddress)) {
```
hedgehog1833, line 135:
```suggestion
    if (currentUser.getEmail().equalsIgnoreCase(emailAddress) || userRepository.existsByEmail(emailAddress)) {
```
I find this really confusing to read here. My suggestion has a little different meaning but makes kind of more sense for me 🤔 Changing the email address to the currently existing one does not really make sense, so why allow it?

hedgehog1833 (approved):

PR #336, opened 2021-02-07
Persist all artist infos from spotify

DanielW1987:
Unfortunately, this PR has become huge.

We now persist all info from Spotify. We don't currently use all of that, but it leaves us room for future use.

In particular, we now also store all images. In the frontend, we can now choose a correct size depending on the specific use case. For example, it doesn't make sense to use the band photos in the largest possible resolution for the search results. This way we save bandwidth above all.


DanielW1987:
[Enrich frontend search results with new infos from Spotify](https://trello.com/c/hCBlZTIw/195-enrich-frontend-search-results-with-new-infos-from-spotify)

DanielW1987:
:tada:

support/src/main/java/rocks/metaldetector/support/ImageBySizeFetcher.java
```diff
@@ -0,0 +1,51 @@
...
+  @ArtifactForFramework
+  @JsonProperty(access = READ_ONLY)
+  default String getThumbnailImage() {
+    return getImageBySizePriority(List.of(XS, S));
+  }
+
+  @ArtifactForFramework
+  @JsonProperty(access = READ_ONLY)
+  default String getSmallImage() {
+    return getImageBySizePriority(List.of(S, M, XS));
```
hedgehog1833, line 29: Order does matter here right? Is 'small' - 'medium' - 'extra small' deliberately in this order? Also for the other methods?

DanielW1987: Yes, the top priority is always the requested quality, then one level better, then one level worse.

hedgehog1833 (commented):
I'm almost sad that I find so few stuff to complain about 😉

spotify/src/test/java/rocks/metaldetector/spotify/facade/dto/ImageSizeTest.java
```diff
@@ -0,0 +1,39 @@
+package rocks.metaldetector.spotify.facade.dto;
```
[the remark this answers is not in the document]
hedgehog1833: Now the package is correct but the class is still in the folder `rocks.metaldetector.spotify.facade.dto`, it has to be moved too.

hedgehog1833 (approved):
🥳

PR #340, opened 2021-02-13
fix releases email if no release is available

DanielW1987:
don't render upcoming or recent releases in email if there is no release available


DanielW1987:
[ReleasesEmail has problems with one empty list](https://trello.com/c/HZiRqGRw/291-releasesemail-has-problems-with-one-empty-list)

hedgehog1833 (approved):

PR #355, opened 2021-03-03
XSS Filter

hedgehog1833:
Here comes the new XSS Filter. It is activated for all Endpoints now. I have no clue if we should remove stuff from the ESAPI.properties file... I always only found these huge examples but as far as I understood we only really need lines 105, 109 and 114.


hedgehog1833:
[Mitigate XSS Attacks](https://trello.com/c/W0d4ugv2/316-mitigate-xss-attacks)

webapp/src/test/java/rocks/metaldetector/security/XSSUtilsTest.java
```diff
@@ -0,0 +1,34 @@
...
+  void test_null_safety() {
+    // when
+    var result = stripXSS(null);
+
+    // then
+    assertThat(result).isNull();
+  }
+
+  @Test
+  @DisplayName("HTML/JS is removed from given String")
```
DanielW1987, line 22:
```suggestion
  @DisplayName("HTML is removed from given String")
```

webapp/src/test/java/rocks/metaldetector/security/XSSUtilsTest.java
```diff
@@ -0,0 +1,34 @@
+package rocks.metaldetector.security;
+
+import org.assertj.core.api.WithAssertions;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import static rocks.metaldetector.security.XSSUtils.stripXSS;
+
+class XSSUtilsTest implements WithAssertions {
```
DanielW1987, line 9: maybe a separate test with javascript input is useful

webapp/src/test/java/rocks/metaldetector/security/XSSFilterTest.java
```diff
@@ -0,0 +1,47 @@
...
+class XSSFilterTest implements WithAssertions {
+
+  private final XSSFilter underTest = new XSSFilter();
+
+  private final MockHttpServletRequest request = new MockHttpServletRequest();
+  private final ServletResponse response = new MockHttpServletResponse();
+  private final FilterChain filterChain = new MockFilterChain();
+
+  @Test
+  @DisplayName("XSSUtils are called to remove bad JS from request body")
```
DanielW1987, line 31: The JavaScript part is confusing, because there is no JavaScript stuff in the request body.

```suggestion
  @DisplayName("XSSUtils are called to sanitize request body")
```

webapp/src/main/resources/ESAPI.properties
```diff
@@ -0,0 +1,545 @@
...
+Authenticator.AllowedLoginAttempts=3
+Authenticator.MaxOldPasswordHashes=13
+Authenticator.UsernameParameterName=username
+Authenticator.PasswordParameterName=password
+# RememberTokenDuration (in days)
+Authenticator.RememberTokenDuration=14
+# Session Timeouts (in minutes)
+Authenticator.IdleTimeoutDuration=20
+Authenticator.AbsoluteTimeoutDuration=120
+
```
DanielW1987, line 92: Which of these properties do we actually use and what happens when you delete all the things we don't use?

What confuses me a lot is, for example, the whole `ESAPI Authenticator` area. Does that have to stay in the file? What happens if you delete that?

Ideally, only the things that have a benefit for us are placed here. If in doubt, I would leave a property in, but I would remove properties without value and things that we have solved elsewhere.

webapp/src/main/resources/ESAPI.properties
```diff
@@ -0,0 +1,545 @@
...
+#
+#===========================================================================
+# ESAPI Configuration
+#
+# If true, then print all the ESAPI properties set here when they are loaded.
+# If false, they are not printed. Useful to reduce output when running JUnit tests.
+# If you need to troubleshoot a properties related problem, turning this on may help.
+# This is 'false' in the src/test/resources/.esapi version. It is 'true' by
+# default for reasons of backward compatibility with earlier ESAPI versions.
+ESAPI.printProperties=true
```
DanielW1987, line 43: This causes some spamming on the console on startup and gives hints about which ESAPI features are currently not configured and can be removed (e.g. validation because `validation.properties` cannot be found - no idea if this is a necessary feature).

webapp/src/main/resources/ESAPI.properties
```diff
@@ -0,0 +1,545 @@
...
+# want to place it in a specific directory.
+Logger.LogFileName=ESAPI_logging_file
+# MaxLogFileSize, the max size (in bytes) of a single log file before it cuts over to a new one (default is 10,000,000)
+Logger.MaxLogFileSize=10000000
+# Determines whether ESAPI should log the user info.
+Logger.UserInfo=true
+# Determines whether ESAPI should log the session id and client IP.
+Logger.ClientInfo=true
+
+#===========================================================================
```
DanielW1987, line 401: this is also very confusing

webapp/src/test/java/rocks/metaldetector/security/XSSRequestWrapperTest.java
```diff
@@ -0,0 +1,211 @@
...
+    // when
+    wrapper.resetInputStream(newContent.getBytes());
+    var result = IOUtils.toString(wrapper.getInputStream(), Charset.defaultCharset());
+
+    // then
+    assertThat(result).isEqualTo(newContent);
+  }
+
+  @Test
+  @DisplayName("XSSUtils are called to remove bad JS from parameter")
```
DanielW1987, line 56:
```suggestion
  @DisplayName("XSSUtils are called to sanitize parameter")
```

webapp/src/test/java/rocks/metaldetector/security/XSSRequestWrapperTest.java
```diff
@@ -0,0 +1,211 @@
...
+      // when
+      var result = new XSSRequestWrapper(request).getParameter(paramName);
+
+      // then
+      assertThat(result).isEqualTo(sanitizedValue);
+    }
+  }
+
+  @Test
+  @DisplayName("XSSUtils are called for every parameter to remove bad JS")
```
DanielW1987, line 92:
```suggestion
  @DisplayName("XSSUtils are called for every parameter to sanitize values")
```

webapp/src/test/java/rocks/metaldetector/security/XSSRequestWrapperTest.java
```diff
@@ -0,0 +1,211 @@
...
+  void test_sanitized_parameters_returned() {
+    // given
+    var paramName = "testParam";
+    var sanitizedValue1 = "sanitizedValue1";
+    var sanitizedValue2 = "sanitizedValue2";
+    request.setParameter(paramName, "badInput1", "badInput2");
+
+    try(MockedStatic<XSSUtils> mock = Mockito.mockStatic(XSSUtils.class)) {
+      mock.when(() -> XSSUtils.stripXSS(any())).thenReturn(sanitizedValue1, sanitizedValue2); // todo hier ist noch was doof
+
```
DanielW1987, line 121: maybe here is noch something doof

webapp/src/test/java/rocks/metaldetector/security/XSSRequestWrapperTest.java
```diff
@@ -0,0 +1,211 @@
...
+      // then
+      assertThat(result.length).isEqualTo(2);
+      assertThat(result[0]).isEqualTo(sanitizedValue1);
+      assertThat(result[1]).isEqualTo(sanitizedValue2);
+    }
+  }
+
+  @Test
+  @DisplayName("XSSUtils are called to remove bad JS from header")
```
DanielW1987, line 133:
```suggestion
  @DisplayName("XSSUtils are called to sanitize header")
```

webapp/src/test/java/rocks/metaldetector/security/XSSRequestWrapperTest.java
```diff
@@ -0,0 +1,211 @@
...
+      // when
+      var result = new XSSRequestWrapper(request).getHeader(headerName);
+
+      // then
+      assertThat(result).isEqualTo(sanitizedValue);
+    }
+  }
+
+  @Test
+  @DisplayName("XSSUtils are called to remove bad JS from headers")
```
DanielW1987, line 169:
```suggestion
  @DisplayName("XSSUtils are called to sanitize headers")
```

webapp/src/main/resources/ESAPI.properties
DanielW1987: just read that we only need lines 105, 109 and 114.

webapp/src/test/java/rocks/metaldetector/security/XSSUtilsTest.java
hedgehog1833: Okay am I just stupid or is injected JS always embedded in a script-block (html) so the test is not really needed as we already have an html test?

webapp/src/test/java/rocks/metaldetector/security/XSSRequestWrapperTest.java
hedgehog1833: I think I just forgot to remove the comment, can't find anything doof 🤔

webapp/src/main/resources/ESAPI.properties
hedgehog1833: I did some cleanups here, the printing is disabled now. Also the part with the validator configuration file is removed. Authentication stuff is commented out. I was not able to remove the logging stuff, it instantly broke then... No matter what I removed.

webapp/src/main/resources/ESAPI.properties
hedgehog1833: Wasn't able to remove that.

webapp/src/test/java/rocks/metaldetector/security/XSSUtilsTest.java
DanielW1987: ...sounds logical.

DanielW1987 (approved):

PR #360, opened 2021-03-05
fix: show releases since today (and not since yesterday)

DanielW1987:
don't know why i used yesterday there :man_shrugging:


DanielW1987:
[Releases page shows releases since yesterday, should be today](https://trello.com/c/NyReWC98/334-releases-page-shows-releases-since-yesterday-should-be-today)

hedgehog1833 (approved):

PR #365, opened 2021-03-08
Following/unfollowing now also possible on homepage

hedgehog1833:
Also fixed a little bug where an artist that is not followed anymore by any users was still shown in the communities favorite artists row. The join statement was wrong there. And yes, this would never be visible in production 😉


hedgehog1833:
[Show follow/unfollow also on start page](https://trello.com/c/0hSNfx0F/300-show-follow-unfollow-also-on-start-page)

DanielW1987 (approved):

PR #369, opened 2021-03-13
Custom remember me service

hedgehog1833:
This service stores the remember me tokens per userId instead of username. I just extended the existing classes and changed only what was necessary. I also wrote tests only for the changes in comparison to the spring classes 😉
The persistent_logins table is truncated once because I was too lazy to update the existing tokens with the correct ids for each user. So everybody has to login again after this is merged 🤷‍♂️
I did not really get your comment regarding the default value and dropping it for in the sql-file in the google-PR. I did the same thing here again, so if you want this another way, you'll have to explain that to me again 😅 My point here is again, that I do not want a default value to exist, but I need it when updating the table (although it is empty at that time). I want an exception to be thrown when somebody tries to insert something with a null-user_id.


hedgehog1833:
[Custom PersistentRememberMeService](https://trello.com/c/6LmVvD1u/338-custom-persistentremembermeservice)

hedgehog1833:
Ah and I will have to make some adjustments after merging  this into the google-PR, because here only the UserEntity is taken into consideration as the AbstractUser does not yet exist.

hedgehog1833:
Closed in favour of https://github.com/MetalDetectorRocks/metal-detector-main/pull/378

PR #370, opened 2021-03-13
gradle cleanup

hedgehog1833:
- dependency management is now working correctly, spring boot version does not have to be added everywhere
- dev-tools are added as developmentOnly on root level (not needed in submodules) so the jar is not packed into the final jar anymore
- deprecated method in root gradle replaced
- gradle wrapper updated

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ buildscript {
 }
 
 plugins {
-  id "java"
+  id "java-library"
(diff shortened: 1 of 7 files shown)


hedgehog1833:
[gradle cleanups](https://trello.com/c/3evBZ3q6/287-gradle-cleanups)

DanielW1987 (approved):

PR #378, opened 2021-03-20
getUsername overridden

hedgehog1833:
The UserEntity's getUsername() now returns the email to achieve having a distinct value in the username-column in the persistent logins table.
I had to add the username as a model in the HomeController as thymeleaf took it by some magic using the getUsername() method, so we could not use this anymore as the email is returned there now.
But all in all it looks better than the custom remember me service I guess.


hedgehog1833:
[Custom PersistentRememberMeService](https://trello.com/c/6LmVvD1u/338-custom-persistentremembermeservice)

hedgehog1833:
If we go this way we should truncate the persistent_logins table, either manually or via flyway-script!

persistence/src/main/java/rocks/metaldetector/persistence/domain/user/UserEntity.java
```diff
@@ -95,6 +95,14 @@ public void setEmail(String newEmail) {
     email = newEmail == null ? "" : newEmail;
   }
 
+  public String getUsername() {
+    return email;
+  }
```
DanielW1987, line 100: I don't like comments in code but maybe here is a little bit explanation meaningful

DanielW1987 (commented):
I think we should clear the table by hand.

DanielW1987:
...looks really good to me 👍

DanielW1987 (approved):

PR #396, opened 2021-04-12
Rest controller for Telegram webhook

hedgehog1833:
I created a controller that receives updates from a webhook. The service takes the message from the update and extracts the email address. Later this will be a code from the frontend like we talked about. But currently this is okay, especially as we can just not register the webhook on prod and nobody is able to do bad stuff. So for testing purposes on preview, this is fine. Then endpoint contains the bot it. This is telegram's idea as kind of a security-by-obscurity thing: the endpoint is public so everybody can call it. If you use the bot id in the endpoint, which only you know, you are kind of sure that it is actually telegram, when it is called.
The `TelegramUpdateServiceImpl` currently just fetches the user and updates his notification config. This service could also do other stuff coming from the webhook, whatever we want to do with the bot.


hedgehog1833:

hedgehog1833:
Before merging this the telegram bot has to be created and the environment variable with the botId set in heroku.

webapp/src/main/java/rocks/metaldetector/service/notification/NotificationConfigTransformer.java
```diff
@@ -12,6 +12,7 @@ public NotificationConfigDto transform(NotificationConfigEntity notificationConf
         .notificationAtReleaseDate(notificationConfigentity.getNotificationAtReleaseDate())
         .notificationAtAnnouncementDate(notificationConfigentity.getNotificationAtAnnouncementDate())
         .frequencyInWeeks(notificationConfigentity.getFrequencyInWeeks())
+        .telegramChatId(notificationConfigentity.getTelegramChatId() == null ? 0 : notificationConfigentity.getTelegramChatId())
```
DanielW1987, line 15: why is 0 cooler than `null` here?

DanielW1987: maybe the chat id should not be primitive?

webapp/src/test/java/rocks/metaldetector/service/telegram/TelegramUpdateServiceImplTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    // when
+    underTest.processUpdate(update);
+
+    // then
+    verify(userRepository).findByEmail(messageText);
+  }
+
+  @Test
+  @DisplayName("is user is present, notificationService is called to set telegram chat id")
```
DanielW1987, line 58:
```suggestion
  @DisplayName("if user is present, notificationService is called to set telegram chat id")
```

webapp/src/main/java/rocks/metaldetector/web/api/request/TelegramChat.java
```diff
@@ -0,0 +1,15 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class TelegramChat {
+
+  private int id;
```
DanielW1987, line 14: Maybe this should be an `Integer`?

webapp/src/main/java/rocks/metaldetector/web/controller/rest/TelegramRestController.java
```diff
@@ -0,0 +1,43 @@
...
+import static rocks.metaldetector.support.Endpoints.Rest.NOTIFICATION_TELEGRAM;
+
+@RestController
+@Slf4j
+public class TelegramRestController {
+
+  private final TelegramUpdateService telegramUpdateService;
+  private final String botId;
+
+  public TelegramRestController(TelegramUpdateService telegramUpdateService, @Value("${telegram.bot-id}") String botId) {
```
DanielW1987, line 27: Can `@Value("${telegram.bot-id}")` not be added at the instance variable level so we can remove the constructor?

webapp/src/main/java/rocks/metaldetector/web/controller/rest/TelegramRestController.java
```diff
@@ -0,0 +1,43 @@
...
+    this.telegramUpdateService = telegramUpdateService;
+    this.botId = botId;
+  }
+
+  @PreAuthorize("permitAll()")
```
DanielW1987, line 32: why is `@PreAuthorize("permitAll()")` needed? 
This is already added in `SecurityConfig`.

webapp/src/test/java/rocks/metaldetector/web/controller/rest/TelegramRestControllerTest.java
```diff
@@ -0,0 +1,128 @@
...
+    return Stream.of(
+        Arguments.of(update.build()),
+        Arguments.of(update.message(
+            TelegramMessage.builder().text("text").build())
+                         .build()),
+        Arguments.of(update.message(
+            TelegramMessage.builder().chat(TelegramChat.builder().build())
+                .build())
+                         .build())
+    );
```
DanielW1987, line 126: the formatting is weird here...
I think an old school constructor fits better here.

webapp/src/main/java/rocks/metaldetector/web/controller/rest/TelegramRestController.java
hedgehog1833: Yeah but then I cannot test it properly because I cannot set the botId without adding a setter..

webapp/src/main/java/rocks/metaldetector/web/controller/rest/TelegramRestController.java
hedgehog1833: Ah yes, forgot to remove that from earlier.

webapp/src/main/java/rocks/metaldetector/web/controller/rest/TelegramRestController.java
DanielW1987: Ah, check!

DanielW1987 (approved):

PR #407, opened 2021-04-24
Telegram notifications

hedgehog1833:
With this PR the existing notifications will also be sent to users who have registered with the telegram bot and therefore a chat id set. The message is formatted very simple, I think the styling can be done in other cards.
What is a bit annoying is that I duplicated some of telegram's api classes as I use them in the webapp module and the telegram module. Maybe an api module would be the right thing but for 2 or 3 classes this seems overengineered to me.


hedgehog1833:

hedgehog1833:
Also not so nice: the part in the `NotificationServiceImpl` whit the `instanceof` thingy to get the correct message for the formatter. I guess there is some refactoring necessary but I wanted to get this to preview at least 😉

discogs/src/main/java/rocks/metaldetector/discogs/config/DiscogsRequestInterceptor.java
```diff
@@ -14,18 +16,16 @@
...
   private final DiscogsConfig discogsConfig;
 
-  DiscogsRequestInterceptor(DiscogsConfig discogsConfig) {
-    this.discogsConfig = discogsConfig;
-  }
-
   @Override
-  public ClientHttpResponse intercept(HttpRequest request, byte[] body, ClientHttpRequestExecution execution) throws IOException {
+  @NonNull
+  public ClientHttpResponse intercept(HttpRequest request, @NonNull byte[] body, ClientHttpRequestExecution execution) throws IOException {
```
DanielW1987, line 28: why do did you introduce `@NonNull` here?

telegram/src/main/java/rocks/metaldetector/telegram/config/TelegramProperties.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.core.env.Environment;
+
+@Configuration
+@PropertySource(value = "classpath:application.yml")
+@ConfigurationProperties(prefix = "telegram")
+@Data
+public class TelegramProperties {
+
+  private final Environment environment;
```
DanielW1987, line 16: Seems like the `environment` is not needed...?

telegram/src/main/java/rocks/metaldetector/telegram/facade/TelegramServiceImpl.java
```diff
@@ -0,0 +1,19 @@
+package rocks.metaldetector.telegram.facade;
+
+import lombok.AllArgsConstructor;
+import org.springframework.stereotype.Component;
+import rocks.metaldetector.telegram.api.TelegramSendMessageRequest;
+import rocks.metaldetector.telegram.client.TelegramClient;
+
+@Component
```
DanielW1987, line 8:
```suggestion
@Service
```

webapp/src/main/java/rocks/metaldetector/security/SecurityConfig.java
```diff
@@ -69,6 +69,7 @@ protected void configure(HttpSecurity http) throws Exception {
         .antMatchers(ACTUATOR_ENDPOINTS).permitAll()
         .antMatchers(Endpoints.Rest.CSP_VIOLATION_REPORT).permitAll()
         .antMatchers(Endpoints.Rest.NOTIFICATION_TELEGRAM + "/" + botId).permitAll()
+        .antMatchers(Endpoints.Rest.NOTIFICATION_TELEGRAM).permitAll()
```
DanielW1987, line 72: Can you please add a ToDo or something like that? That's not supposed to be a permanent solution, is it?

webapp/src/main/java/rocks/metaldetector/service/notification/NotificationServiceImpl.java
```diff
@@ -134,8 +139,13 @@ private void frequencyNotification(NotificationConfigEntity notificationConfigEn
...
 
-          notificationConfigEntity.setLastNotificationDate(now);
-          notificationConfigRepository.save(notificationConfigEntity);
+          if (notificationConfig.getTelegramChatId() != null) {
+            String message = notificationFormatter.formatFrequencyNotificationMessage(upcomingReleases, recentReleases);
+            telegramService.sendMessage(notificationConfig.getTelegramChatId(), message);
+          }
+
+          notificationConfig.setLastNotificationDate(now);
+          notificationConfigRepository.save(notificationConfig);
```
DanielW1987, line 148: I don't like the fact that Telegram and e-mail are mixed here. It would probably be better to have a separate service for e-mail and Telegram, which would then be used by the NotificationService. Don't know if this will work. The class is blowing up quite a bit right now.

In addition, the whole thing only seems to work if I have activated the e-mail notifications. This should work independently of each other.

discogs/src/main/java/rocks/metaldetector/discogs/config/DiscogsRequestInterceptor.java
hedgehog1833: Because I got warnings from idea otherwise 🤷‍♂️

webapp/src/main/java/rocks/metaldetector/security/SecurityConfig.java
hedgehog1833: Oooh I just forgot to remove that! Good catch :) That was for testing purposes so that I could trigger notifications from postman locally. Is removed now.

webapp/src/main/java/rocks/metaldetector/service/notification/NotificationServiceImpl.java
hedgehog1833: Yeah I thought so too but I'm not quite sure how to separate here in a smart way because just removing the new 4 lines will not solve the problem of this class being huge... I think we should talk about this this week some time.

hedgehog1833: And yeah the part with telegram and email being separate from each other needs some more thought too. The telegram notifications are currently "extra" and the notifications on release/announcement date don't look like they also work with telegram.

hedgehog1833: I'm currently working on the UI for the telegram registration I guess I will work out something there..

DanielW1987 (approved):

PR #408, opened 2021-04-25
Fixed a bug with too many genres when synchronizing from spotify

hedgehog1833:


DanielW1987 (approved):
you could have implemented this here so that the genres are not just hard truncated, but that is finely separated at the comma, so you don't have broken genres in the database. I also don't see a test for this.

PR #413, opened 2021-05-05
Telegram configuration possible per UI

hedgehog1833:
This PR brings the Telegram UI. On the notification settings page a user can generate a registration ID. This id is sent by the user to our bot and the webhook saves it.
Still open: your comments from the last PR regarding the `NotificationService`. I did split it up into `NotificationService` and `NotificationConfigService` but still I'd like to do some work on the `NotificationService`. But in the next PR because this one has already over 20 files. That next step also includes decoupling the telegram notifications from the email notifications.
Another nice to have thing would be a success message from the bot when registering has succeeded.


hedgehog1833:

webapp/src/main/resources/templates/frontend/notification-settings.html
```diff
@@ -50,6 +50,23 @@ <h2 class="h5 mt-3">Extra email notifications</h2>
...
                 </div>
             </fieldset>
+
+            <h2 class="h5 mt-3">Telegram notifications</h2>
+            <hr class="border-secondary"/>
+            <fieldset id="telegram-area">
+                <div id="telegram-activation-area">
+                    <p>Notifications can be sent per Telegram too! To activate this generate a registration ID and send it in your Telegram app to our bot 'metal_detector_notification_bot'.</p>
+                    <div class="form-group">
+                        <button class="btn btn-outline-dark" id="telegram-id-button">Generate registration ID</button>
```
DanielW1987, line 60:
```suggestion
                        <button class="btn btn-secondary" id="telegram-id-button">Generate registration ID</button>
```

webapp/src/test/java/rocks/metaldetector/service/notification/NotificationServiceImplTest.java
```diff
@@ -12,6 +12,7 @@
 import org.mockito.Captor;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
+import org.mockito.MockedStatic;
```
DanielW1987, line 15: this class has some unused imports

DanielW1987 (commented):
Some questions:

- why disappears the registration id in the frontend if I reload the page?
- do we also get some kind of confirmation in Telegram that it worked? That would make sense.
- Is it possible to test the whole thing locally?

hedgehog1833:
Regarding your questions:
- Didn't think about your first point, so this has to be fixed, yes
- I wanted to put the confirmation message in another PR, but as it is just a one liner I can add it here
- Testing locally is not that easy because you cannot let the webhook be sent to your local machine. But if you register at preview you can fetch the correct chat id and then set it in your local DB. This is how I did it. Remember that currently you need to send your email to the bot until this PR here is merged

hedgehog1833:
I'll take a look at your other comments later, guess I'm outside today 🌞

DanielW1987 (approved):

PR #418, opened 2021-05-15
Rework of NotificationConfigEntity

hedgehog1833:
I'm really sorry for this huge PR but cutting this smaller would have been a pain 😅
Like discussed we now have a `NotificationConfigEntity` and a `TelegramConfigEntity`. Having a NotificationConfig is now not mandatory for a user anymore because it is created on the fly when updating via the frontend for the first time. Fetching the configs now returns a new api object that contains all configs separated per channel. Updating the different configs works more or less the same as before per single config. Channel specific info like the telegram registration id get their own controller to separate them from the normal configs which are the same for every channel.
I also refactored some repositories on the way. The never have to explicitly use entities ids when the entity is used as a foreign key, you can always use the entity itself. That simplifies a lot of calls and especially tests.
The NotificationService still works the same as before. I'm refactoring that one already on a separate branch based on this PR but that is another quite big change for later (but looks already pretty nice 👌)


hedgehog1833:
[Rework NotificationConfig](https://trello.com/c/yFV4qbPU/357-rework-notificationconfig)

hedgehog1833:
Will reopen with more stuff 💪

hedgehog1833:
[Rework NotificationConfig](https://trello.com/c/yFV4qbPU/357-rework-notificationconfig)

PR #419, opened 2021-05-15
Rework of Notifications-Backend

hedgehog1833:
I'm really sorry for this huge PR but cutting this smaller would have been a pain 😅
Like discussed we now have a `NotificationConfigEntity` and a `TelegramConfigEntity`. Having a NotificationConfig is not mandatory for a user anymore because it is created on the fly when updating via the frontend for the first time. Fetching the configs now returns a new api object that contains all configs separated per channel. Updating the different configs works more or less the same as before per single config. Channel-specific info like the telegram registration id get their own controller to separate them from the normal configs which are the same for every channel.
I also refactored some repositories on the way. They never have to explicitly use entities ids when the entity is used as a foreign key. You can always use the entity itself. That simplifies a lot of calls and especially tests.
The `NotificationService` is completely remade and split up into a Scheduler, ReleaseCollector and a Sender interface that has to be implemented for every channel. This change also leads to the butler being queried for every notification config instead of once per method call but makes it way more readable and extendable.


hedgehog1833:
[Rework NotificationConfig](https://trello.com/c/yFV4qbPU/357-rework-notificationconfig)

hedgehog1833:
[Rework NotificationConfig](https://trello.com/c/yFV4qbPU/357-rework-notificationconfig)

webapp/src/main/java/rocks/metaldetector/service/notification/config/NotificationConfigServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    AbstractUserEntity currentUser = currentUserSupplier.get();
+
+    NotificationChannel channel = NotificationChannel.getChannelFromString(notificationConfigDto.getChannel());
+    Optional<NotificationConfigEntity> notificationConfigOptional = notificationConfigRepository.findByUserAndChannel(currentUser, channel);
+    NotificationConfigEntity notificationConfig;
+
+    notificationConfig = notificationConfigOptional.orElseGet(() -> NotificationConfigEntity.builder()
+        .user(currentUser)
+        .channel(channel)
+        .build());
```
DanielW1987, line 48:
```suggestion
    NotificationConfigEntity notificationConfig = notificationConfigOptional.orElseGet(() -> NotificationConfigEntity.builder()
        .user(currentUser)
        .channel(channel)
        .build());
```

webapp/src/main/java/rocks/metaldetector/service/notification/config/NotificationConfigServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+    notificationConfig.setFrequencyInWeeks(notificationConfigDto.getFrequencyInWeeks());
+    notificationConfig.setNotificationAtAnnouncementDate(notificationConfigDto.isNotificationAtAnnouncementDate());
+    notificationConfig.setNotificationAtReleaseDate(notificationConfigDto.isNotificationAtReleaseDate());
+    notificationConfig.setNotify(notificationConfigDto.isNotify());
```
DanielW1987, line 53: you may want to add this to transformer to remove mapping from business logic?!

webapp/src/main/java/rocks/metaldetector/service/notification/config/TelegramConfigServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+      log.warn("Could not set telegram chat id for registration id '{}' - id not found", registrationId);
+    }
+  }
+
+  @Override
+  @Transactional
+  public int generateRegistrationId() {
+    int registrationId;
+    do { registrationId = ThreadLocalRandom.current().nextInt(100_000, 1_000_000); }
+    while (telegramConfigRepository.existsByRegistrationId(registrationId));
```
DanielW1987, line 69: I am thinking about whether we should add some kind of endless loop protection here? Maybe a termination after 100 runs with logging?

DanielW1987: And yes, I know I've already approved it before ;)

webapp/src/main/java/rocks/metaldetector/service/notification/messaging/EmailNotificationSender.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+  }
+
+  @Override
+  public void sendAnnouncementDateMessage(AbstractUserEntity user, List<ReleaseDto> todaysAnnouncements) {
+    if (!todaysAnnouncements.isEmpty()) {
+      emailService.sendEmail(new TodaysAnnouncementsEmail(user.getEmail(), user.getUsername(), todaysAnnouncements));
+    }
+  }
+}
```
DanielW1987, line 40: I really like that refactoring.

webapp/src/main/java/rocks/metaldetector/service/notification/messaging/NotificationScheduler.java
```diff
@@ -0,0 +1,73 @@
...
+
+@Component
+@AllArgsConstructor
+public class NotificationScheduler {
+
+  private final NotificationConfigRepository notificationConfigRepository;
+  private final NotificationReleaseCollector notificationReleaseCollector;
+  private final NotificationSenderSupplier notificationSenderSupplier;
+
+  @Scheduled(cron = "0 0 4 * * SUN")
```
DanielW1987, line 24: should we align this to 7?

```suggestion
  @Scheduled(cron = "0 0 7 * * SUN")
```

webapp/src/main/java/rocks/metaldetector/service/notification/messaging/NotificationScheduler.java
```diff
@@ -0,0 +1,73 @@
...
+  @Transactional
```
DanielW1987, line 25:
```suggestion
  @Transactional(readOnly = true)
```

webapp/src/main/java/rocks/metaldetector/service/notification/messaging/NotificationSenderSupplier.java
```diff
@@ -0,0 +1,32 @@
...
+public class NotificationSenderSupplier implements Function<NotificationChannel, NotificationSender> {
+
+  private final NotificationSender emailService;
+  private final NotificationSender telegramService;
+
+  public NotificationSenderSupplier(@Qualifier("emailNotificationSender") NotificationSender emailService,
+                                    @Qualifier("telegramNotificationSender") NotificationSender telegramService) {
+    this.emailService = emailService;
+    this.telegramService = telegramService;
+  }
```
DanielW1987, line 19: What happen if you name the instance variables `emailNotificationSender`? Maybe you don't need the constructor...

webapp/src/main/java/rocks/metaldetector/service/notification/messaging/TelegramNotificationSender.java
```diff
@@ -0,0 +1,67 @@
...
+  static final String TODAYS_ANNOUNCEMENTS_TEXT = "Today's metal release announcements:";
+
+  private final TelegramMessagingService telegramMessagingService;
+  private final TelegramConfigRepository telegramConfigRepository;
+  private final TelegramNotificationFormatter telegramNotificationFormatter;
+
+  @Override
+  @Transactional(readOnly = true)
+  public void sendFrequencyMessage(AbstractUserEntity user, List<ReleaseDto> upcomingReleases, List<ReleaseDto> recentReleases) {
+    if (!(upcomingReleases.isEmpty() && recentReleases.isEmpty())) {
```
DanielW1987, line 29: Perhaps it would be better to check whether the list is empty in the scheduler? The check is currently redundant across all channels.

webapp/src/main/resources/db/migration/v7_0__notification_configs.sql
```diff
@@ -0,0 +1,17 @@
...
+create table telegram_configs(
+    id bigserial not null constraint telegram_configs_pkey primary key,
+    created_by varchar(255),
+    created_date timestamp,
+    last_modified_by varchar(255),
+    last_modified_date timestamp,
+    chat_id integer,
+    registration_id integer,
+    notification_configs_id bigint not null constraint fktbume816qt0bpad3goy3gw8aq references notification_configs
+);
```
DanielW1987, line 17: maybe you shold merge this with version 5 and 6?

You have to check if column exist:

```sql
ALTER TABLE tableName
DROP COLUMN IF EXISTS columnName;
```

DanielW1987 (commented):
Really like that...
just small things

webapp/src/main/java/rocks/metaldetector/service/notification/messaging/NotificationScheduler.java
hedgehog1833: This one is not readonly, the notification config is updated with the lastnotificationdate

webapp/src/main/java/rocks/metaldetector/service/notification/messaging/TelegramNotificationSender.java
hedgehog1833: Yep, looks better.

webapp/src/main/java/rocks/metaldetector/service/notification/config/NotificationConfigServiceImpl.java
hedgehog1833: hmm this would be a weird transformer as it takes an entity and a dto as parameters right? 🤔  Because the dto does not have all the info (like the user) that the entity has...

webapp/src/main/java/rocks/metaldetector/service/notification/config/TelegramConfigServiceImpl.java
hedgehog1833: Yeah that may be smart...

webapp/src/main/resources/db/migration/v7_0__notification_configs.sql
hedgehog1833: When I merge the flyway scripts I get migration failures 🤔

DanielW1987: As discussed....leave it as it is.

DanielW1987 (approved):

DanielW1987:
Will merge it myself right away to continue with the UI.

PR #422, opened 2021-05-22
Genre now has a maximum of 255 characters, test added

hedgehog1833:
Apparently that bug wasn't fixed correctly last time, there was one character too much. I added a unit test this time 😉

### DIFF
--- a/webapp/src/test/java/rocks/metaldetector/service/artist/transformer/ArtistEntityTransformerTest.java
+++ b/webapp/src/test/java/rocks/metaldetector/service/artist/transformer/ArtistEntityTransformerTest.java
@@ -5,6 +5,8 @@ import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
 import java.util.Collections;
+import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


webapp/src/test/java/rocks/metaldetector/service/artist/transformer/ArtistEntityTransformerTest.java
```diff
@@ -77,4 +79,19 @@ void should_transform_discogs_artist() {
     assertThat(artistEntity.getImageM()).isEqualTo(discogsArtist.getImages().get(M));
     assertThat(artistEntity.getImageL()).isEqualTo(discogsArtist.getImages().get(L));
   }
+
+  @Test
+  @DisplayName("the genre has a maximum of 255 characters")
+  void test_genre_max_length() {
+    // given
+    var genres = IntStream.range(1, 70).mapToObj(String::valueOf).collect(Collectors.toList());
```
DanielW1987, line 87: Maybe something like 

```java
"Black Metal".repeat(50);
```

would be more understandable.

DanielW1987 (approved):

PR #439, opened 2021-06-13
 Spring Boot OAuth Integration with Spotify

hedgehog1833:
Having a bit more experience with OAuth in Spring Boot now, I refactored parts of the Spotify Authentication stuff. The normal Artist-Search uses the client_credentials-OAuth-Flow. This one is pretty straight forward and can be easily taken care of by spring boot.
I introduced an OAuth-Interceptor that uses the spring boot integration and is executed with every search request. Here is the one downside and reason this is currently a draft: the token is not cached anymore at the moment. I'm still thinking about a nice solution here.
Another thing I'm not completely sure about is putting this in the support package, as now every module has the dependency to spring-security. But the OAuth-Interceptor can be used in different modules sooner or later so 🤷‍♂️ 
I guess the User-Authorization with Spotify can also be done by spring boot and we can get rid of the SpotifyAuthorizationClient completely, but first I want the search to be nice.


hedgehog1833:
It got a bit more complicated than anticipated but I think the current solution is fine now.
The support module provides an Interceptor  and a TokenSupplier. The interceptor just sets the header value into the request, the TokenSupplier does the actual requests to fetch the token. 
What's special about the TokenSupplier is that it's a bean with scope "protoype", which means that every time it is autowired anywhere, a new instance of this class is created. This is necessary because each Supplier has its own registrationId, a value that is set in the application.yml in the oauth configuration. Creating beans dynamically with values from the application.yml is possible, but also quite complicated so I just went the easy way of adding a setter for the registrationId and set the value while configuring the rest template.
So far so good, now the next issue was caching. Logically we need separate caches for each module (or each api) that uses this interceptor. This means, that just adding the @Cacheable annotation to the TokenSupplierImpl does only work as long as the Interceptor is only used once. To tackle this problem, each module that wants to use the interceptor with caching needs to implement a wrapper class, which implements the TokenSupplier-interface. It autowires the "Default-Impl" and just passes the calls. This wrapper now gets the @Cacheable annotation and specifies the cache to use.

hedgehog1833:
The funny thing is that I started with the plan to get rid of stuff and ended up with: +472 −408 😅

support/src/main/java/rocks/metaldetector/support/oauth/OAuth2AccessTokenSupplier.java
```diff
@@ -0,0 +1,8 @@
+package rocks.metaldetector.support.oauth;
+
+import java.util.function.Supplier;
+
+public interface OAuth2AccessTokenSupplier extends Supplier<String> {
+
+  void setRegistrationId(String registrationId);
+}
```
DanielW1987, line 8: I don't think that's a good thing. The interface inherits from the functional interface `Supplier`, but it will not keep the functional interface behaviour. 

`get()` vs. `setRegistrationdId()` is also a bit confusing.

spotify/src/main/java/rocks/metaldetector/spotify/config/SpotifyOAuth2AccessTokenSupplierWrapper.java
```diff
@@ -0,0 +1,24 @@
...
+  @Cacheable("spotifyAppAuthorizationToken")
+  public String get() {
+    return oAuth2AccessTokenSupplier.get();
+  }
+
+  @Override
+  public void setRegistrationId(String registrationId) {
+    oAuth2AccessTokenSupplier.setRegistrationId(registrationId);
+  }
+}
```
DanielW1987, line 24: The class has a dependency on an interface that the class itself implements?
This looks very strange and I wonder if it can't be done more simply. Let's talk about it on the phone.

DanielW1987: Maybe inheritance works better here?

support/src/main/java/rocks/metaldetector/support/oauth/OAuth2ClientManagerConfig.java
```diff
@@ -0,0 +1,30 @@
...
+        .build();
+    var manager = new AuthorizedClientServiceOAuth2AuthorizedClientManager(clients, oAuth2AuthorizedClientService);
+    manager.setAuthorizedClientProvider(authorizedClientProvider);
+    return manager;
+  }
+
+  @Bean
+  public String registrationId() {
+    return "spotify";
+  }
```
DanielW1987, line 29: Don't understand why you need this bean here?

NilsDammenhayn: Me neither, deleted it.

spotify/src/main/java/rocks/metaldetector/spotify/config/SpotifyOAuth2AccessTokenSupplierWrapper.java
NilsDammenhayn: What we needed was the `JdbcOAuth2AuthorizedClientService`, providing this as a bean does all the magic (with the correct table existing)

DanielW1987 (approved):
Like that.

PR #457, opened 2021-07-23
Spotify user login with Spring Boots OAuth integration

hedgehog1833:
As discussed for google users the spotify sync does currently not work. I'm not 100% sure if the config in the application-preview.yml is necessary..


hedgehog1833:
[Rework of Spotify Authorization](https://trello.com/c/K4zYIgm5/382-rework-of-spotify-authorization)

spotify/src/main/java/rocks/metaldetector/spotify/client/SpotifyUserLibraryClientImpl.java
```diff
@@ -85,11 +77,10 @@ public SpotifyFollowedArtistsPage fetchFollowedArtists(String token, String next
     return result.getArtistsPage();
   }
 
-  private HttpEntity<Object> createHttpEntity(String token) {
+  private HttpEntity<Object> createHttpEntity() {
```
DanielW1987, line 80: If the token is already set in an interceptor, can't the accept and charset be set there as well? Seems unnecessarily elaborate to me here.

webapp/src/main/java/rocks/metaldetector/service/user/events/UserDeletionEventListener.java
```diff
@@ -42,7 +40,6 @@ public void onApplicationEvent(UserDeletionEvent event) {
     AbstractUserEntity user = event.getUserEntity();
     log.info("User '" + user.getPublicId() + "' deleted");
 
-    spotifyAuthorizationRepository.deleteByUser(user);
```
DanielW1987, line 45: What happens to the record in the OAuth table when a user deletes himself?

webapp/src/main/resources/application-preview.yml
```diff
@@ -21,6 +21,12 @@ spring:
           auth: true
           starttls:
             enable: true
+  security:
+    oauth2:
+      client:
+        registration:
+          spotify-app:
+            redirect-uri: "https://{baseHost}{basePort}{basePath}/settings/spotify-synchronization"
```
DanielW1987, line 29: where is `baseHost`, `basePort` and `basePath` defined?

DanielW1987: How do the values differ from the `application.yml`? I don't think we need them. But not sure.

DanielW1987 (commented):
Looks good

webapp/src/main/java/rocks/metaldetector/service/user/events/UserDeletionEventListener.java
DanielW1987: ok, this is already fixed.

DanielW1987 (approved):
Maybe you can address these issues in your current work

spotify/src/main/java/rocks/metaldetector/spotify/client/SpotifyUserLibraryClientImpl.java
hedgehog1833: We do that in multiple clients, I created a card for it.

webapp/src/main/resources/application-preview.yml
hedgehog1833: I think somehow spring does not yet know (at least from an application.yml perspective) that ssl is activated on preview and prod. So probably we can refactor some more here, but I'll check that out later. Don't want to break more than necessary.

PR #462, opened 2021-07-31
Fixed Spotify sync for OAuth Users; OAuth Config and user handling improved

hedgehog1833:
The creation of OAuth Users is moved to the spot, where the user's are loaded. It is now also easy to add more OAuth Login provider like spotify.


hedgehog1833:
[Improve OAuth User handling](https://trello.com/c/Xy0S4FjX/386-improve-oauth-user-handling)

webapp/src/main/resources/static/ts/src/clients/oauth-rest-client.ts
```diff
@@ -18,15 +17,19 @@ export class OauthRestClient {
...
             })
             .catch((error: AxiosError) => {
-                this.toastService.createErrorToast(UNKNOWN_ERROR_MESSAGE);
-                throw error;
+                if (error.response?.status == 404) {
+                    return 404;
+                } else {
+                    this.toastService.createErrorToast(UNKNOWN_ERROR_MESSAGE);
+                    throw error;
+                }
```
DanielW1987, line 32: It would be better if a Boolean comes back here, so that is not so low-level in the service with status code evaluation.

DanielW1987 (approved):
Looks good.

PR #475, opened 2021-08-29
Fix duplicates in notifications, reissue notification configuration

hedgehog1833:
The mail/telegram notifications now only contain releases with state "OK". Also reissues are only returned if the notification config is configured that way.


hedgehog1833:
[Releases E-Mail also contains duplicates and maybe also faulty releases.](https://trello.com/c/dZfMWxWp/343-releases-e-mail-also-contains-duplicates-and-maybe-also-faulty-releases)

hedgehog1833:
[Don't notify for reissues](https://trello.com/c/Pz34wdlT/352-dont-notify-for-reissues)

DanielW1987 (approved):

PR #476, opened 2021-08-29
Get and Post header interceptors introduced, logging interceptor + tests

hedgehog1833:
I added two new interceptors for get and post requests. Also a new LoggingInterceptor exists now that is also properly tested with nice lib to test logging.


hedgehog1833:
[Set Header per Interceptor](https://trello.com/c/QAT1hENy/384-set-header-per-interceptor)

DanielW1987 (approved):

PR #481, opened 2021-09-04
Improve performance for fetching most expected releases

hedgehog1833:
I rewrote the database query to fetch the number of followers directly with the TopArtists instead of querying followers for every single artist. Also the `collectTopFollowedArtists()` method was called twice.


hedgehog1833:
[Start page loads to slow (>5 sec)](https://trello.com/c/2HBtxJRG/387-start-page-loads-to-slow-5-sec)

DanielW1987 (approved):

PR #493, opened 2021-09-07
spotify logging

DanielW1987:
On local machine I have the following output during spotify connect:

```
2021-09-07 15:36:06,506 [http-nio-8080-exec-10] INFO  r.m.c.l.SpotifyRequestLoggingFilter - GET /settings/spotify-synchronization, client=0:0:0:0:0:0:0:1, session=B4C588DA7FC7873FFD637993C4DA637F, user=Administrator] 
2021-09-07 15:36:06,580 [http-nio-8080-exec-2] INFO  r.m.c.logging.RestRequestLoggingFilter - GET /rest/v1/oauth/spotify-user, client=0:0:0:0:0:0:0:1, session=B4C588DA7FC7873FFD637993C4DA637F, user=Administrator] 
```

The one with query parameter is not logged as it is not handled by one of our mvc controller. 

So, in prod we may should see the query params there.

### DIFF
--- a/webapp/src/main/java/rocks/metaldetector/config/logging/SpotifyRequestLoggingFilter.java
+++ b/webapp/src/main/java/rocks/metaldetector/config/logging/SpotifyRequestLoggingFilter.java
@@ -1,0 +1,37 @@
+package rocks.metaldetector.config.logging;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import org.springframework.web.filter.CommonsRequestLoggingFilter;
+import rocks.metaldetector.support.Endpoints;
+import rocks.metaldetector.support.infrastructure.WithSensitiveDataRemover;
+
+import javax.servlet.http.HttpServletRequest;
+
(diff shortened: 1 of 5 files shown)


webapp/src/main/java/rocks/metaldetector/config/logging/SpotifyRequestLoggingFilter.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+public class SpotifyRequestLoggingFilter extends CommonsRequestLoggingFilter implements WithSensitiveDataRemover {
+
+  SpotifyRequestLoggingFilter() {
+    super.setIncludeQueryString(true);
+    super.setIncludePayload(true);
+    super.setIncludeClientInfo(true);
+    super.setMaxPayloadLength(10000);
+    super.setIncludeHeaders(false);
```
hedgehog1833, line 17: Maybe set this to true too?

hedgehog1833 (approved):

PR #498, opened 2021-09-11
Remove csp

hedgehog1833:


hedgehog1833:
[Remove csp stuff from spring boot application and use nginx for that](https://trello.com/c/pHw5Vngt/392-remove-csp-stuff-from-spring-boot-application-and-use-nginx-for-that)

DanielW1987 (approved):

PR #502, opened 2021-09-17
Fix 404 error log in client when spotify auth was not found

hedgehog1833:


hedgehog1833:
[do not log missing spotify auth as error in client](https://trello.com/c/J4MmrOgu/388-do-not-log-missing-spotify-auth-as-error-in-client)

DanielW1987 (approved):

PR #507, opened 2021-09-23
Customize oauth login request to get refresh token from google

hedgehog1833:
Google doesn't send a refresh token by default, these additional parameters are needed. So now "remember me" works, as the refresh token is automatically used to get a new token if the old one expired.


hedgehog1833:
[Remember-me for OAuth User](https://trello.com/c/wy1J8ICm/344-remember-me-for-oauth-user)

DanielW1987 (approved):

PR #509, opened 2021-09-23
reponsive release page

DanielW1987:
Here it is, finally.
Would be good if you could test something too.


DanielW1987:
[Responsive Release Page](https://trello.com/c/FMy75L77/359-responsive-release-page)

hedgehog1833 (approved):
This looks amazing ❤️ 
Just a few things that can also be done later:
- The menu on the left (releases, blog etc.) and the filter menu on the releases page seem to work differently. It is possible to remove the new filter menu per click next to the menu but not the old one. This could be unified.
- I think we should have a "back to defaults" button for the filters.
- On my local testing I didn't see images on my phone, but I guess this is something with my local setup 🤔

PR #523, opened 2021-11-06
Users with expired tokens are cleaned up every night 

hedgehog1833:
I also build an endpoint to trigger this from the admin area but I had some weird behaviors locally where it seemed like the wrong user was deleted (in admin area **and** database) until I restarted the application. This did not occur when the scheduler ran the job. I will test this a bit more before adding the endpoint. 
Currently the jobs runs once per night.


hedgehog1833:
[Cleanup users with expired tokens](https://trello.com/c/WlFZUkGZ/407-cleanup-users-with-expired-tokens)

hedgehog1833:
Found the issue: my query was not correct. I joined users to tokens but had to join tokens to users. Now everything also works fine with the endpoint triggered from the admin area.

webapp/src/test/java/rocks/metaldetector/web/controller/rest/CleanupRestControllerIT.java
```diff
@@ -0,0 +1,47 @@
...
+    @DisplayName("Administrator is allowed to POST on endpoint " + CLEANUP + "'")
+    @WithMockUser(roles = "ADMINISTRATOR")
+    void admin_is_allowed_to_cleanup() throws Exception {
+      mockMvc.perform(post(CLEANUP))
+              .andExpect(status().isOk());
+    }
+  }
+
+  @Nested
+  @DisplayName("Users is not allowed to send requests to all endpoints")
```
DanielW1987, line 36:
```suggestion
  @DisplayName("User is not allowed to send requests to all endpoints")
```

DanielW1987 (commented):
Maybe only cleanup is a little bit to generic.

What do you think about 'registration cleanup'?

persistence/src/main/java/rocks/metaldetector/persistence/domain/user/UserRepository.java
```diff
@@ -21,4 +22,7 @@
 
   @Query("select case when count(u) > 0 then true else false end from users u where dtype = 'native_users' and u.username = :username")
   boolean existsByUsername(@Param("username")String username);
+
+  @Query(value = "select * from users as u left join tokens as t on u.id = t.users_id where t.expiration_date_time < NOW() and t.token_type = 'EMAIL_VERIFICATION'", nativeQuery = true)
```
DanielW1987, line 26: why not an inner join?

hedgehog1833: Good question, fixed it.

DanielW1987 (approved):

PR #535, opened 2021-11-24
Fix ci job for dependabot

hedgehog1833:
The first ci run in dependabot PRs currently always fails. This fixes that :)


DanielW1987 (approved):
Were there any updates to GitHub or why was this necessary?

PR #569, opened 2022-01-10
Fixes the usage of oauth clients for scheduled tasks

hedgehog1833:
This turned out to be way more work than I first thought 😓 
Context: since we are using oauth, the notifications are not working anymore. There was an Exception in the ButlerInterceptor because the authentication is null. That is because scheduled tasks are not authenticated. To fix that I now manually set an anonymous user for the scheduled notification tasks. Then everything failed because Spring Security needs a ServletRequest to fetch a new authentication token (for whatever reasons) which also is not available in a task that is scheduled by Spring. To fix that I needed to add the new implementation of the OAuth2AuthorizedClientManager which was designed specifically for this purpose and only has to be used for these scheduled requests. To achieve that, I check if the current thread is scheduled by checking the thread name, which we prefix in the Config of the TaskScheduler. I don't know if there is a more elegant way to find out if a task is scheduled, but this works.

### DIFF
--- a/support/src/main/java/rocks/metaldetector/support/infrastructure/ApacheHttpClientConfig.java
+++ b/support/src/main/java/rocks/metaldetector/support/infrastructure/ApacheHttpClientConfig.java
@@ -17,7 +17,6 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.scheduling.TaskScheduler;
-import org.springframework.scheduling.annotation.EnableScheduling;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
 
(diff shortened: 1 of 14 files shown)


support/src/main/java/rocks/metaldetector/support/infrastructure/ApacheHttpClientConfig.java
```diff
@@ -26,10 +25,10 @@
 import static java.util.concurrent.TimeUnit.MINUTES;
 
 @Configuration
-@EnableScheduling
 @Slf4j
 public class ApacheHttpClientConfig {
 
+  public static final String SCHEDULED_TASK_NAME_PREFIX = "threadPoolTaskScheduler";
```
DanielW1987, line 31: you have ignored the LucaNet indentation :D

webapp/src/main/java/rocks/metaldetector/service/notification/messaging/NotificationReleaseCollector.java
```diff
@@ -34,12 +36,10 @@
 
     if (!followedArtistNames.isEmpty()) {
       var now = LocalDate.now();
-      upcomingReleases = releaseService.findAllReleases(followedArtistNames, new TimeRange(now, now.plusWeeks(frequency))).stream()
-          .filter(release -> release.getState().equals(RELEASE_STATE_OK))
+      upcomingReleases = findReleases(followedArtistNames, new TimeRange(now, now.plusWeeks(frequency))).stream()
```
DanielW1987, line 39: I don't understand why the check for the release state was thrown out here. With this we potentially have duplicates in the notifications again.

hedgehog1833: Because I call the User-Endpoint on the Butler now. Before the admin endpoint was called and that one gives all releases. The user endpoint only gives state ok!

DanielW1987: fetchimage

DanielW1987 (approved):

PR #570, opened 2022-01-11
Add new envs to docker compose

hedgehog1833:
This is a bit annoying... One more reason to have a frontend to administrate the clients.


DanielW1987 (approved):
ah sorry, its the docker-compose.

PR #607, opened 2022-02-23
pin version for auth service

DanielW1987:


hedgehog1833:
Maybe you should add an tag for the butler image too? There is also no latest anymore

PR #625, opened 2022-03-05
Remove TokenEntity and everything related

hedgehog1833:
I removed the TokenEntity and everything related as discussed. Please take a look at the new logic for removing expired users: I now check if there were ever any changes to the user (created_date = last_modified_date), if the user is still disabled (should be redundant but to be sure 🤷‍♂️) and if it is older than 10 days (the lifetime of the token).


persistence/src/main/java/rocks/metaldetector/persistence/domain/user/UserRepository.java
```diff
@@ -21,8 +21,10 @@
   boolean existsByEmail(String email);
 
   @Query("select case when count(u) > 0 then true else false end from users u where dtype = 'native_users' and u.username = :username")
-  boolean existsByUsername(@Param("username")String username);
+  boolean existsByUsername(@Param("username") String username);
 
-  @Query(value = "select * from users as u inner join tokens as t on u.id = t.users_id where t.expiration_date_time < NOW() and t.token_type = 'EMAIL_VERIFICATION'", nativeQuery = true)
-  List<AbstractUserEntity> findAllWithExpiredToken();
+  // todo NilsD umbauen auf nicht-aktiviert und erstellungsdatum
+//  @Query(value = "select * from users as u inner join tokens as t on u.id = t.users_id where t.expiration_date_time < NOW() and t.token_type = 'EMAIL_VERIFICATION'", nativeQuery = true)
```
DanielW1987, line 27: what about this todo?

webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
```diff
@@ -172,55 +170,27 @@ public UserDto getCurrentUser() {
...
       throw new TokenExpiredException();
     }
 
-    UserEntity userEntity = tokenEntity.getUser();
+    // get user from token
+    UserEntity userEntity = (UserEntity) userRepository.findByPublicId(claims.getSubject())
+        .orElseThrow(() -> new ResourceNotFoundException(USER_WITH_ID_NOT_FOUND.toDisplayString()));
 
-    // 4. set new password
+    // set new password
```
DanielW1987, line 191: Doesn't look quite round to me yet, that in `TokenServiceImpl` the registration token is checked and in `UserServiceImpl` the password reset token. 

This should happen at least in one service. I don't really care in which one, probably in `UserServiceImpl`. 

Maybe you could also build an own method to extract the user from the token, something like:

```java
private UserEntity extractUserFromToken(String tokenString) {
  var claims = jwtsSupport.getClaims(tokenString);

  if (claims.getExpiration().before(new Date())) {
      throw new TokenExpiredException();
  }

  return (UserEntity) userRepository.findByPublicId(claims.getSubject())
        .orElseThrow(() -> new ResourceNotFoundException(USER_WITH_ID_NOT_FOUND.toDisplayString()));
}
```

webapp/src/test/java/rocks/metaldetector/service/cleanup/RegistrationCleanupServiceImplTest.java
```diff
@@ -1,112 +1,92 @@
...
+//  private NotificationConfigRepository notificationConfigRepository;
+//
+//  @InjectMocks
+//  private RegistrationCleanupServiceImpl underTest;
+//
+//  @AfterEach
+//  void tearDown() {
+//    reset(userRepository, notificationConfigRepository);
+//  }
+//
```
DanielW1987, line 40: Do we still need this test class?

persistence/src/main/java/rocks/metaldetector/persistence/domain/user/UserRepository.java
hedgehog1833: Done, just forgot to remove comment and old line.

webapp/src/test/java/rocks/metaldetector/service/cleanup/RegistrationCleanupServiceImplTest.java
hedgehog1833: Jep, no clue why I commented this... Guess I just forgot about it.

webapp/src/main/java/rocks/metaldetector/service/user/UserServiceImpl.java
hedgehog1833: I moved the `verifyEmailToken` method back to the UserService to have the check only there. The TokenService still needs the jwtsSupport to get the public user id, but I think that's okay as it is a slightly different use case.

DanielW1987 (approved):

PR #704, opened 2022-05-21
Update rest-assured and thymeleaf dialect (groovy 4)

hedgehog1833:


build.gradle.kts
```diff
@@ -1,22 +1,20 @@
...
-  "org.junit" to libs.versions.junit.get()
+    "org.springframework" to libs.versions.spring.get(),
+    "org.springframework.security" to libs.versions.springSecurity.get(),
+    "org.springframework.boot" to libs.versions.springBoot.get(),
+    "com.fasterxml.jackson.core" to libs.versions.jackson.get(),
+    "com.fasterxml.jackson.datatype" to libs.versions.jackson.get(),
+    "com.fasterxml.jackson.module" to libs.versions.jackson.get(),
+    "org.mockito" to libs.versions.mockito.get(),
+    "org.junit.jupiter" to libs.versions.junit.get(),
+    "org.apache.groovy" to "4.0.2"
```
DanielW1987, line 17: Why do we need Groovy here instead of junit?

hedgehog1833: "org.junit" was junit4, which we don't use anymore. We need groovy here because rest-assured and the thymeleaf dialect use different version of groovy 4 (4.0.0 and 4.0.1). So deleting org.junit was a cleanup and adding groovy was necesssary to update the two dependencies.

hedgehog1833: org.junit.jupiter (=junit5) is still in there in line 16

DanielW1987 (approved):

PR #710, opened 2022-05-25
Refactored deprecated config

hedgehog1833:
The WebSecurityConfigurerAdapter is deprecated so I refactored the config and did some cleanups.


hedgehog1833:
Forgot to look at tests, this will take some time 😓

hedgehog1833:
It looks like annotations like SpringBootTest or WebMvcTest and also Import need to be on the test classes and cannot be on any class they extend from 🤷‍♂️ Also the HttpSecurity Bean must not be mocked anymore as our SecurityConfig now works a bit differently.

hedgehog1833:
Also I renamed the integration test classes from ...IT to ...IntegrationTest as this seems to be the naming convention. Idea complained about ...IT for a while now.

DanielW1987 (approved):

PR #741, opened 2022-06-25
Error handling refined

hedgehog1833:
Just like in https://github.com/MetalDetectorRocks/metal-release-butler/pull/254 I improved the status codes on our rest endpoints. I did a quick test in the frontend and everything looked fine.

### DIFF
--- a/webapp/src/main/java/rocks/metaldetector/service/exceptions/RestExceptionsHandler.java
+++ b/webapp/src/main/java/rocks/metaldetector/service/exceptions/RestExceptionsHandler.java
@@ -20,6 +20,7 @@ import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.client.RestClientException;
 import org.springframework.web.context.request.ServletWebRequest;
 import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.method.annotation.MethodArgumentTypeMismatchException;
 import org.springframework.web.servlet.ModelAndView;
 import rocks.metaldetector.config.constants.ViewNames;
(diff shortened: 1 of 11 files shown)


webapp/src/main/java/rocks/metaldetector/service/exceptions/RestExceptionsHandler.java
```diff
@@ -36,45 +37,39 @@
...
 import static org.springframework.http.HttpStatus.NOT_FOUND;
+import static org.springframework.http.HttpStatus.SERVICE_UNAVAILABLE;
 import static org.springframework.http.HttpStatus.UNPROCESSABLE_ENTITY;
 import static org.springframework.http.HttpStatus.UNSUPPORTED_MEDIA_TYPE;
 
 @ControllerAdvice
 @Slf4j
 public class RestExceptionsHandler {
 
-  @ExceptionHandler({IllegalArgumentException.class, MissingServletRequestParameterException.class})
```
DanielW1987, line 46: why you removed the `IllegalArgumentException`

DanielW1987 (approved):

webapp/src/main/java/rocks/metaldetector/service/exceptions/RestExceptionsHandler.java
hedgehog1833: Because it is not needed anymore. The only place it could be thrown was in the Controller where the Source Enum was built which his is now handled by spring (and throws the MethodArgumentTypeMismatchException). From an api-wise perspective the IllegalArgumentException always feels a bit weird because it is so broad. Usually a more specific exception can be found which is better to be handled in the RestExceptionsHandler.

PR #826, opened 2022-09-04
Merge jwt login

hedgehog1833:


webapp/src/main/java/rocks/metaldetector/security/JwtAuthenticationFilter.java
```diff
@@ -0,0 +1,64 @@
...
+  protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
+    try {
+      String token = parseHeader(request);
+      if (token != null) {
+        Claims claims = jwtsSupport.getClaims(token);
+        AbstractUserEntity user = userRepository.findByPublicId(claims.getSubject())
+            .orElseThrow(() -> new ResourceNotFoundException(USER_WITH_ID_NOT_FOUND.toDisplayString()));
+
+        UsernamePasswordAuthenticationToken authentication = new UsernamePasswordAuthenticationToken(user, null, user.getAuthorities());
+        authentication.setDetails(authenticationDetailsSource.buildDetails(request));
```
DanielW1987, line 44: why not passing the `WebAuthenticationDetailsSource` directly here and removing the bean?

```suggestion
        authentication.setDetails(new WebAuthenticationDetailsSource().buildDetails(request));
```

webapp/src/main/java/rocks/metaldetector/security/SecurityConfig.java
```diff
@@ -69,14 +66,17 @@ public class SecurityConfig {
...
+  private final BCryptPasswordEncoder bCryptPasswordEncoder;
 
   @Value("${telegram.bot-id}")
   private String botId;
 
   @Bean
   SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
     http
-      .csrf().ignoringAntMatchers(REST_ENDPOINTS, ACTUATOR_ENDPOINTS)
+      .csrf().ignoringAntMatchers(Endpoints.Rest.LOGIN, ACTUATOR_ENDPOINTS)
```
DanielW1987, line 79: I think you can remove the other login route to avoid constant naming conflicts here.

webapp/src/main/java/rocks/metaldetector/security/SecurityConfig.java
```diff
@@ -136,13 +138,21 @@ SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
...
-        .accessDeniedHandler(new CustomAccessDeniedHandler(() -> SecurityContextHolder.getContext().getAuthentication()))
-        .defaultAuthenticationEntryPointFor(new LoginUrlAuthenticationEntryPoint(LOGIN), new AntPathRequestMatcher(HOME))
-        .defaultAuthenticationEntryPointFor(new HttpStatusEntryPoint(UNAUTHORIZED), new AntPathRequestMatcher(REST_ENDPOINTS));
+//      .exceptionHandling()
+//        .accessDeniedHandler(new CustomAccessDeniedHandler(() -> SecurityContextHolder.getContext().getAuthentication()))
+//        .defaultAuthenticationEntryPointFor(new LoginUrlAuthenticationEntryPoint(LOGIN), new AntPathRequestMatcher(HOME))
+//        .defaultAuthenticationEntryPointFor(new HttpStatusEntryPoint(UNAUTHORIZED), new AntPathRequestMatcher(REST_ENDPOINTS));
+        .exceptionHandling().authenticationEntryPoint(authenticationEntryPoint)
+      .and()
+        .addFilterBefore(authenticationFilter, UsernamePasswordAuthenticationFilter.class);
```
DanielW1987, line 147: Can you please also add the following:

```java
.sessionManagement().sessionCreationPolicy(SessionCreationPolicy.STATELESS)
```

support/src/test/java/rocks/metaldetector/support/infrastructure/CustomClientErrorHandlerTest.java
```diff
@@ -51,7 +47,7 @@ void has_error_should_return_true(ClientHttpResponse response) throws IOExceptio
     boolean hasError = underTest.hasError(response);
 
     // then
-    assertThat(hasError).isTrue();
+    assertTrue(hasError);
```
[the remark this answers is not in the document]
hedgehog1833: I knew you would say this 😉 I did this because the testing library I used for testing the mocked logger uses the same method signatures as assertj. Which meas I have to use an import there. I'll change it :)

webapp/src/main/resources/application.yml
```diff
@@ -143,3 +143,6 @@ log-csp-report: false
 telegram:
   rest-base-url: https://api.telegram.org
   bot-id: ${TELEGRAM_BOT_ID}
+
+frontend:
+  origin: http://localhost:3000
```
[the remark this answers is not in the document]
hedgehog1833: Currently we always redirect to `www.metal-detector.rocks`, I guess so www is need here too? Or change the redirection back to the non-www-version (did not have the effect I wanted anyway when I did it).

webapp/src/main/java/rocks/metaldetector/security/JwtAuthenticationFilter.java
hedgehog1833: Hm why? Can't test it as nice then and a new instance that has to be garbage collected at some point is created on every server interaction.

webapp/src/main/java/rocks/metaldetector/security/SecurityConfig.java
hedgehog1833: That would be the route to the oauth login page 🤔 We still need that^^

DanielW1987: ok, right. In the future, however, this must be done differently with Google OAuth. Currently, the old login route also contains HTML that is delivered by the server.

support/src/test/java/rocks/metaldetector/support/infrastructure/CustomClientErrorHandlerTest.java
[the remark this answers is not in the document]
DanielW1987: Ah ok, I see...However, the verify assertions from before had done a good job, right?

webapp/src/main/resources/application.yml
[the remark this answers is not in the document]
DanielW1987: Ah, ok...think we need the www

Not so easy to change that because the redirection is made with 301 (move **permanently**).

webapp/src/main/java/rocks/metaldetector/security/SecurityConfig.java
hedgehog1833: ah right.... yeah later :D

PR #844, opened 2022-09-18
Login cleanups

DanielW1987:


webapp/src/main/java/rocks/metaldetector/security/SecurityConfig.java
```diff
@@ -75,7 +74,6 @@ public class SecurityConfig {
   private final OAuth2AuthorizationRequestResolver oAuth2AuthorizationRequestResolver;
   private final JwtAuthenticationEntryPoint authenticationEntryPoint;
   private final JwtAuthenticationFilter authenticationFilter;
-  private final BCryptPasswordEncoder bCryptPasswordEncoder;
```
hedgehog1833, line 78: We have to check this out later. I kind of remember that without this bean on registration something does not work... But everything is so long ago that I don't remember when we removed this from being configured 🙈

hedgehog1833 (approved):
lgtm 👍

PR #897, opened 2022-10-30
return refresh token on sign in

DanielW1987:
A refresh token is now returned as a cookie when logging in. 

The refresh tokens are persisted and de facto replace the Remember Me functionality. 

On logout, the Refresh Token is deleted from the database (not yet implemented). The Access Token is only valid for 15 minutes anyway.

Next steps:
- Cleanup of expired Refresh Tokens
- Access Token Refresh with Refresh Token
- Logout


webapp/src/main/java/rocks/metaldetector/service/user/RefreshTokenService.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private ResponseCookie createCookie(String refreshToken) {
+    return ResponseCookie.from(COOKIE_NAME, refreshToken)
+        .maxAge(Duration.ofMinutes(securityProperties.getRefreshTokenExpirationInMin() - OFFSET_IN_MINUTES))
+        .secure(securityProperties.isSecureCookie())
+        .httpOnly(true)
+        .path("/")
+        .domain(domain)
+        .sameSite("Strict")
```
hedgehog1833, line 49: I think we should use 'Lax' here. Can't really explain now but remember to have read something about strict beeing too strict sometimes 😉

webapp/src/main/resources/db/migration/v13_0__drop_persistent_logins.sql
```diff
@@ -0,0 +1,4 @@
+-- Creation Date: 2022-10-30
+-- Description: remove persistent_logins, because remember me is implemented with refresh tokens
+
+drop table persistent_logins;
```
hedgehog1833, line 4: The `UserDeletionEventListener` deletes from this table when a user account is deleted. I think this isn't removed in this PR.

hedgehog1833 (commented):
I didn't go to deep into the unit tests, trust you there :D still two questions

webapp/src/main/java/rocks/metaldetector/service/user/RefreshTokenService.java
DanielW1987: Ok, but let's test first with Strict, since frontend and backend under the same domain

webapp/src/main/resources/db/migration/v13_0__drop_persistent_logins.sql
DanielW1987: ok, will have a look

MetalDetectorRocks (approved):

PR #905, opened 2022-11-09
add and cleanup some integration tests to prepare for a later pr

DanielW1987:
I've updated all the rest controller security tests. From the setup, I also had to adjust something, because that no longer ran in my other branch.


webapp/src/test/java/rocks/metaldetector/web/controller/rest/UserAccountRestControllerIntegrationTest.java
```diff
@@ -0,0 +1,177 @@
...
+          )
+          .andExpect(status().isOk());
+    }
+  }
+
+  @Nested
+  class UserRoleTest {
+
+    @Test
+    @DisplayName("User is allowed to GET on endpoint " + USERS + "' (all users)")
```
hedgehog1833, line 94:
```suggestion
    @DisplayName("User is not allowed to GET on endpoint " + USERS + "' (all users)")
```

See also other tests in this class for role user.

webapp/src/test/java/rocks/metaldetector/web/controller/rest/UserAccountRestControllerIntegrationTest.java
```diff
@@ -0,0 +1,177 @@
...
+    @WithMockUser
+    void user_is_allowed_to_fetch_all_users() throws Exception {
```
hedgehog1833, line 96:
```suggestion
    void user_is_not_allowed_to_fetch_all_users() throws Exception {
```

hedgehog1833 (approved):
only a few test names, looks good otherwise 👍

PR #941, opened 2022-12-11
Refresh tokens

DanielW1987:
Next steps:

- [ ] Cleanup of expired Refresh Tokens
- [x] Access Token Refresh with Refresh Token
- [ ] Logout


webapp/src/main/java/rocks/metaldetector/service/auth/TokenPair.java
```diff
@@ -0,0 +1,9 @@
+package rocks.metaldetector.service.auth;
+
+import org.springframework.http.ResponseCookie;
+
+public record TokenPair(
+    String accessToken,
+    ResponseCookie refreshToken
+) {
+}
```
hedgehog1833, line 9:
```suggestion
) {}
```
Like in AccessTokenResponse

hedgehog1833 (approved):
👍

PR #959, opened 2022-12-30
move from auth service to spring filter and handler

DanielW1987:
essentially, there are only two things that have changed:

- Removed classes from the MVC world regarding login
- Switch to filter & handler instead of own auth service


webapp/src/main/java/rocks/metaldetector/security/filter/CustomUsernamePasswordAuthenticationFilter.java
```diff
@@ -0,0 +1,37 @@
...
+  public CustomUsernamePasswordAuthenticationFilter(
+      AuthenticationConfiguration authenticationConfiguration,
+      ObjectMapper objectMapper
+  ) throws Exception {
+    super(authenticationConfiguration.getAuthenticationManager());
+    this.objectMapper = objectMapper;
+  }
+
+  @Override
+  @SneakyThrows
```
hedgehog1833, line 28: Why did you use that?

webapp/src/test/java/rocks/metaldetector/security/filter/CustomAuthorizationFilterTest.java
```diff
@@ -35,7 +35,7 @@
 import static org.springframework.http.HttpHeaders.AUTHORIZATION;
 
 @ExtendWith(MockitoExtension.class)
-class JwtAuthenticationFilterTest implements WithAssertions {
+class CustomAuthorizationFilterTest implements WithAssertions {
```
hedgehog1833, line 38: Here a test could be added that FORBIDDEN is sent in case of an exception.

webapp/src/test/java/rocks/metaldetector/security/filter/CustomUsernamePasswordAuthenticationFilterTest.java
```diff
@@ -0,0 +1,111 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.verify;
+
+@ExtendWith(MockitoExtension.class)
+class CustomUsernamePasswordAuthenticationFilterTest implements WithAssertions {
```
hedgehog1833, line 33: Maybe one verify() for the object mapper in one of the tests?

hedgehog1833 (approved):

webapp/src/main/java/rocks/metaldetector/security/filter/CustomUsernamePasswordAuthenticationFilter.java
DanielW1987: `SneakyThrows` converts checked exception to runtime exception at compile time. The `RestExceptionHandler` can than do the rest of the work.

`objectMapper.readValue()` throws checked exception.

PR #966, opened 2023-01-07
Handle logout

DanielW1987:


webapp/src/main/java/rocks/metaldetector/security/handler/CustomLogoutHandler.java
```diff
@@ -0,0 +1,29 @@
...
+public class CustomLogoutHandler implements LogoutHandler {
+
+  private final RefreshTokenService refreshTokenService;
+
+  @Override
+  public void logout(HttpServletRequest request, HttpServletResponse response, Authentication authentication) {
+    Stream.of(request.getCookies())
+        .filter(cookie -> REFRESH_TOKEN_COOKIE_NAME.equals(cookie.getName()))
+        .map(Cookie::getValue)
+        .forEach(refreshTokenService::removeRefreshToken);
```
hedgehog1833, line 27: I think we don't need this Handler at all, because you configured the `deleteCookies()` method in the `SecurityConfg`. At least the JavaDoc reads like it 🤔

webapp/src/test/java/rocks/metaldetector/security/handler/CustomLogoutHandlerTest.java
```diff
@@ -0,0 +1,41 @@
...
+    Cookie cookie = new Cookie(REFRESH_TOKEN_COOKIE_NAME, tokenCookieValue);
+    doReturn(new Cookie[]{cookie}).when(request).getCookies();
+
+    // when
+    underTest.logout(request, null, null);
+
+    // then
+    verify(refreshTokenService).removeRefreshToken(tokenCookieValue);
+  }
+}
```
hedgehog1833, line 41:
```suggestion

}
```

hedgehog1833: If this class is needed at all, see other comment.

hedgehog1833 (approved):

PR #973, opened 2023-01-14
Registration verification email prep

DanielW1987:
I converted the abstract Email class to an interface and stripped out the logic a bit. I think that it is not necessary that we also show the URLs for the preview imprint or privacy policy site in emails that are sent on preview environment. That makes the implementation a bit more convenient to understand.


webapp/src/main/java/rocks/metaldetector/service/email/EmailServiceImpl.java
```diff
@@ -17,17 +18,18 @@
...
 @Slf4j
-@AllArgsConstructor
+@RequiredArgsConstructor
 public class EmailServiceImpl implements EmailService {
 
   private final JavaMailSender emailSender;
   private final ISpringTemplateEngine templateEngine;
   private final MailProperties mailProperties;
   private final MimeMessageHelperFunction messageHelperFunction;
+  private String frontendBaseUrl;
```
hedgehog1833, line 28:
```suggestion
  @Setter
  @Value("${frontend.origin}")
  private String frontendBaseUrl;
```
So the method below is not needed?

hedgehog1833 (approved):

webapp/src/main/java/rocks/metaldetector/service/email/EmailServiceImpl.java
DanielW1987: nice, will do in follow-up pr.

PR #1028, opened 2023-03-06
account details refactorings

DanielW1987:


webapp/src/main/java/rocks/metaldetector/config/WebMvcConfig.java
```diff
@@ -25,47 +19,17 @@ public class WebMvcConfig implements WebMvcConfigurer {
   @Value("${frontend.origin}")
   private String frontendOrigin;
 
-  @Override
-  public void configurePathMatch(PathMatchConfigurer configurer) {
-    configurer.setUseTrailingSlashMatch(true);
```
hedgehog1833, line 30: Not sure anymore why I added this on migrating to Spring Boot 3 but if everything works fine without it, I'm good :)

hedgehog1833 (approved):
lgtm

webapp/src/main/java/rocks/metaldetector/config/WebMvcConfig.java
DanielW1987: I guess this is only relevant in the context of MVC.

PR #1101, opened 2023-04-22
wrap getUserByEmailOrUsername into transation

DanielW1987:


MetalDetectorRocks (approved):

DanielW1987:
[Forgot password with username leads to 500](https://trello.com/c/8gdG8Yn5/435-forgot-password-with-username-leads-to-500)

PR #1167, opened 2023-06-04
Admin dashboard API

hedgehog1833:
Added a rest controller with an API for the dashboard data in the admin area. Data is currently mocked and will be added one by one.


DanielW1987 (approved):
Fine for me but I'm not really happy with the naming `AdminDashboard` because it reflects the frontend a bit too much.

Better I would find:

- `/api/v1/statistics`
- `/api/v1/metrics`

or maybe something other in this way.

PR #1252, opened 2023-08-12
remove hibernate version enforcement; onetoone to manytoone

hedgehog1833:
as already said i'm not entirely sure why this has to be like this now. but hibernate 6.2 seems to have [changed something](https://docs.jboss.org/hibernate/orm/6.2/migration-guide/migration-guide.html#logical-1-1-unique) and this works 🤷‍♂️


DanielW1987:
@hedgehog1833 Let me also check that on my machine before merging.

persistence/src/main/java/rocks/metaldetector/persistence/domain/artist/FollowActionEntity.java
```diff
@@ -23,12 +23,12 @@
...
+  @ManyToOne(targetEntity = AbstractUserEntity.class)
   @JoinColumn(nullable = false, name = "user_id")
   @NonNull
   private AbstractUserEntity user;
 
-  @OneToOne(targetEntity = ArtistEntity.class)
+  @ManyToOne(targetEntity = ArtistEntity.class)
   @JoinColumn(nullable = false, name = "artist_id")
   @NonNull
   private ArtistEntity artist;
```
DanielW1987, line 34: `OneToOne` was definitely wrong before.

persistence/src/main/java/rocks/metaldetector/persistence/domain/notification/NotificationConfigEntity.java
```diff
@@ -26,7 +26,7 @@
 @Entity(name = "notificationConfigs")
 public class NotificationConfigEntity extends BaseEntity {
 
-  @OneToOne(targetEntity = AbstractUserEntity.class)
+  @ManyToOne(targetEntity = AbstractUserEntity.class)
```
DanielW1987, line 29: I think here `OneToOne` is correct, because a user can only have 1 notification config and not more and the newly created unique constraint should not cause a problem.

persistence/src/main/java/rocks/metaldetector/persistence/domain/notification/TelegramConfigEntity.java
```diff
@@ -21,7 +21,7 @@
 @Entity(name = "telegramConfigs")
 public class TelegramConfigEntity extends BaseEntity {
 
-  @OneToOne(targetEntity = NotificationConfigEntity.class)
+  @ManyToOne(targetEntity = NotificationConfigEntity.class)
```
DanielW1987, line 24: I think here `OneToOne` is correct, because a user can only have 1 notification config and not more and the newly created unique constraint should not cause a problem.

DanielW1987: @hedgehog1833
do you have any problems/warnings if this stays at `ManyToMany`?

hedgehog1833: @DanielW1987 The only way our integration tests don't fail is currently with OneToOne. Thing is that I am not entirely sure if we have to change something else so that ManyToOne works again 🤷‍♂️

hedgehog1833: And with this point in the [hibernate migration guide](https://docs.jboss.org/hibernate/orm/6.2/migration-guide/migration-guide.html#logical-1-1-unique) is kind of sounds right? 😅

DanielW1987 (approved):

PR #1254, opened 2023-08-14
Statistics: Follow Infos

hedgehog1833:
Old branch with lots of commits and rebasing stuff going on 😅 I'll squash that. Removes mock data for following infos in the Statistics endpoint and adds the real stuff.


webapp/src/main/java/rocks/metaldetector/service/statistics/StatisticsServiceImpl.java
```diff
@@ -48,4 +54,17 @@ private UserInfo buildUserInfo() {
...
+    List<FollowActionEntity> allFollowActions = followActionRepository.findAll();
+    Map<YearMonth, Long> followActionsPerMonth = allFollowActions.stream()
+        .collect(Collectors.groupingBy(followActionEntity -> YearMonth.of(followActionEntity.getCreatedDateTime().getYear(), followActionEntity.getCreatedDateTime().getMonth()),
+                                       TreeMap::new,
+                                       Collectors.counting()));
+    return ArtistFollowingInfo.builder()
+        .followingsPerMonth(followActionsPerMonth)
+        .totalFollowings(allFollowActions.size())
+        .followingsThisMonth(followActionsPerMonth.getOrDefault(YearMonth.now(), 0L))
+        .build();
```
DanielW1987, line 68: Have you checked if it is possible to do some pre-aggregation at database level?

DanielW1987 (approved):

webapp/src/main/java/rocks/metaldetector/service/statistics/StatisticsServiceImpl.java
hedgehog1833: No but as I am currently doing this for the release infos, I'll check on that here too!

hedgehog1833: Refactored to use a database query to get the info. Long time since I had to do java 8 stream stuff 😄

DanielW1987 (approved):

PR #1323, opened 2023-10-23
Update java to 21

hedgehog1833:


DanielW1987 (approved):
So far so good, but please merge only if you have a good feeling...

PR #1377, opened 2024-02-11
OAuth Authorization Code Flow with Spotify

hedgehog1833:
I have to give some context here so this will be some text:
I wanted to build the flow here just like it was, just without an http session (stateless so to speak) so I had to jump through some loops.
1. In the frontend I still use `window.location.href` to initiate the flow because when I switched to axios I get CORS erros. Unsure if a problem with my code or Spotify or maybe it is just like that. So href it is. Now when doing so it is not possible to send any headers, so from the backend's perspective the call is anonymous. To solve that I save the value I would put in the header on an axios call inside a Cookie.
2. That cookie is read in the `OAuth2AuthorizationCodeSaveRequestFilter` which saves the current user with a generated state to the database. This exact state is then also used for Spotify. To achieve that the `OAuth2AuthorizationCodeStateGenerator` is scoped per request. It generates one state per request. Later in the `OAuth2AuthorizationRequestResolver` the state in the request is overwritten with the value generate here (in `OAuth2AuthorizationCodeSaveRequestFilter`.
3. Spotify redirects to the backend and the `OAuth2AuthorizationCodeLoginFilter` checks for the state provided by Spotify (the one we sent in 2.). If found, the user in the database is loaded and authenticated. This leads to Spring's `OAuth2AuthorizationCodeGrantFilter` then being able to persist the authorized client and the user is forwarded to the Spotify configuration page.
What I am not sure about: if the cookie is the way to go here or if there is some other way. More on that in the frontend pull request. But changing this would only mean a different way of obtaining the current user in `OAuth2AuthorizationCodeSaveRequestFilter` so hopefully no big deal.


hedgehog1833:
@DanielW1987 This is ready for review now. I now use the existing cookie instead of creating a new one.

DanielW1987 (approved):

